PR #25, opened 2020-07-15
DockerFile + add tutorial

sofianelounici:
For the easy install (the UI), we need to keep the WORKDIR command otherwise the .html templates cannot be found in Flask.

I also added a Jupyter Notebook containing a tutorial explaining the Basic Usage / Advanced Usage. The Advanced Usage part is still in progress


sofianelounici:
> Why have you written this tutorial? It's almost same as the wiki page. Maybe it's better just to use the wiki so that users don't have 3 different sources of information (readme+tutorial+wiki)

For the Basic Usage yes, for the Advanced Usage, you can run the notebook and see the results (the Wiki is just description). For Python users, it is always nice to work with Jupyter Notebooks

sofianelounici:
> We don't have integrated the similarity model. This should be done in the UI, which needs some effort because as it is now it's not very nice

Yes it should be integrated (not the goal of my request though)

tutorials/01 - Getting Started with Credential Digger.ipynb
```diff
@@ -0,0 +1,328 @@
...
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "import numpy as np\n",
+    "\n",
+    "from credentialdigger.cli import Client\n",
+    "\n",
+    "c = Client(dbname='DB', dbuser='1234',\n",
+    "                          dbpassword='1234',\n",
+    "                          dbhost='localhost', dbport=5432)"
```
marcorosa, line 159: wrong indentation

tutorials/01 - Getting Started with Credential Digger.ipynb
```diff
@@ -0,0 +1,328 @@
...
+      "Number of leaks:  605\n",
+      "Number of false positives: 0\n"
+     ]
+    }
+   ],
+   "source": [
+    "url = 'https://github.com/Mebus/cupp'\n",
+    "discoveries = c.scan(repo_url=url,\n",
+    "                          models=[],\n",
+    "                          verbose=False)\n",
```
marcorosa, line 215: wrong indentation

sofianelounici: Done

marcorosa:
Sorry, I forgot to merge 😅😂

Issue #3, opened 2020-06-17
Improve user interface


marcorosa:
The ui needs several improvements.
- ~Remove the "fake path" placeholder and use the real path when uploading a file with rules~ [#9 ]
- ~Fix the button to show/hide the false positives~
- ~Fix the scan function (it always uses all the categories)~ [#35 and #39 ]
  - ~Even if the checkbox is not checked, all the rules are selected~ [#35]
- ~Fix the button to re-scan a repository (choose the category instead of picking rules, like in normal scans)~ [#44]
- ~Add flag to force repo rescan (see #36 )~ [#44]

alaabenfatma:
The first point has been taken care of here: https://github.com/SAP/credential-digger/pull/9

alaabenfatma:
> 
>     * Fix the scan function (it always uses all the categories)
>       
>       * Even if the checkbox is not checked, all the rules are selected
Does this mean that the scan operation is always looking for leaks that belong to all the existing categories? I have failed to spot this issue while tackling with the UI. That said, I have created this temporary repo (https://github.com/alaabenfatma/cdigger_foo/) that contains 5 leaks : 
- 5 passwords
- 1 token
Scanning this repo (from the UI) while selecting only the 'token' category gives a correct result, please see screenshot below:
![image](https://user-images.githubusercontent.com/9027148/91254858-b3a5fd80-e763-11ea-9b38-7a6afe7eb370.png)

marcorosa:
Hi, that was a note on the behaviour before your PR #35 . You solved that bug where all the rules were always used even if the checkbox was not checked.

Now, I did a quick test, and I noted another bug:
if you select a category, and you also check the checkbox to use all the rules, only the rules of the category are used to scan the repository. In theory, I expect to choose the rules "last clicked".

- On the UI point of view, the selection of one category vs the "all rules" checkbox should be mutually exclusive (better UX)
- On the scanner point of view, if the checkbox is checked then all the rules should be used (checkbox wins against the selection of a single category, but in this case the single category should not be highlighted)

marcorosa:
All issues raised have been fixed in v2.0.0

PR #30, opened 2020-08-13
Fixed show/hide FPs button

alaabenfatma:
# Overview
The toggle button that displays/hides the false positive has been fixed. (Targetted issue: https://github.com/SAP/credential-digger/issues/3)

### Notes
- This solution performs better than the one that has been proposed earlier since we are not retrieving the list of discoveries every time we filter out the FPs.
- This solution is not the fastest. It can be easily optimised in order to achieve further speed gains by introducing an array that contains the false positives the moment the window loads. We currently are traversing the array of discoveries every time we want to filter the FPs, which is not optimal. I will optimize it later. 
- This approach does not refresh the page, all the changes happen in-place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![Logo](github_assets/Logo-CD-Mint_48.png)
+![Logo](https://raw.githubusercontent.com/SAP/credential-digger/master/github_assets/Logo-CD-Mint_48.png)
 
 # Credential Digger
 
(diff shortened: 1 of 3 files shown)


alaabenfatma:
> 
> 
> Hi, thanks for this fix!
> Please change the base branch to `develop`, then I will merge it

Hi @marcorosa, thank you for the review.
Please feel free to merge whenever you see fit.

PR #39, opened 2020-08-26
Minor fix: Alternate between the states of the checkbox and the category selector.

alaabenfatma:
### Targetted issues
https://github.com/SAP/credential-digger/issues/3#issuecomment-680747770

#### Overview
This PR helps enhance the UX.

- Activating the 'Use all rules' checkbox clears the selections made on the 'Categories selector'
- Selecting an item from the 'Categories selector' disables the 'Use all rules' checkbox if it is active.
- Selecting no category at all && Not activating the checkbox will stop you from starting a scan.

### DIFF
--- a/ui/res/js/repos.js
+++ b/ui/res/js/repos.js
@@ -35,13 +35,22 @@ document.getElementById('startRepoScan').addEventListener('click', function () {
   // show loading popup
 });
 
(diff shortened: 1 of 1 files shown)


marcorosa (approved):
Great, thanks

Issue #37, opened 2020-08-26
flagging discoveries disable hide FP


marcorosa:
When the *Hide FP* button is activated but the user flags a false positive with the *False positive* button, the Hide FP gets disabled

marcorosa:
Fixed in #40

PR #40, opened 2020-08-26
Fix: Keep filtering FPs after refreshing the page

alaabenfatma:
#### Targetted issue
https://github.com/SAP/credential-digger/issues/37

#### Overview
The FPs will get filtered even after the page gets refreshed. The problem was that the changes happen out-of-place and require the page to reload, hence losing all the settings. This would, in theory, have been avoided if we were using AJAX instead, or at least be less complicated to deal with for the changes happening actually in-place.

#### Fix
I have made use of the __window.sessionStorage__ in order to store the state of the flag (whether to filter the FPs or not) next time the window loads.

#### Limitation
A latency between loading the discoveries and filtering them may be noticed.

#### Future work
Should we reset the filtering setting back to default (filter = false) when the window closes?

### DIFF
--- a/ui/res/js/discoveries.js
+++ b/ui/res/js/discoveries.js
@@ -3,6 +3,17 @@
 //var requests = [];
 var filterFPs = false;
 
+// It is mandatory to wait for the window to load before executing the
+// scripts 
+window.onload = function () {
+  filterFPs = sessionStorage.getItem('filterFPs');
+  if (filterFPs == 'None') {
+    filterFPs = false;
+  }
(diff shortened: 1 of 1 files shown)


marcorosa:
*Should we reset the filtering setting back to default (filter = false) when the window closes?*
I don't have an opinion on this... I would expect to open the page with the discoveries of a repository and seeing also the false positive, that I can later hide with the button. But it's just my personal feeling.

marcorosa (approved):

PR #41, opened 2020-08-26
Improve default rules

ichbinfrog:
**Combined private key rules into single rule**
```yaml
  - definition: -----BEGIN [A-Z]+ PRIVATE KEY( BLOCK)?-----
    category: crypto_key
    description: private key
```
This reduces the cost of matching the `[\r\n]+(?:\w+:.+)*[\s]*(?:[0-9a-zA-Z+\/=]{64,76}[\r\n]+)+[0-9a-zA-Z+\/=]+[\r\n]+` group (in most scenarios the beginning of a private key block would automatically be followed by the corresponding key). The trade-off between cost and precision can be further discussed.

**Removed generic rules**
```yaml
  - regex: sshpass|password|pwd|passwd|pass
    category: password
    description: password keywords
```
This rule often leads to a lot of false positive due to it being completely context insensitive, for example;
- `passed` is matched because it contains `pass` 
- in a key-value file, it would even match when the value is empty 
These kind of rules should rather be delegated to the models and not regexes.

```yaml
  - regex: ((?:\?|\&|\"|\')(?:access_token)(?:\"|\')?\s*(?:=|:))
    category: access token
    description: generic access token
```
The same can be said about the above rule. It lacks the context for proper matching.

**Added some more distinct API key matching rules**
```yaml
  - definition: 'xox.-[0-9]{12}-[0-9]{12}-[0-9a-zA-Z]{24}'
    category: token
    description: Slack API key

  - definition: 'access_token,production$[0-9a-z]{161[0-9a,]{32}'
    category: token
    description: paypal acess token

  - definition: 'amzn.mws]{8}-[0-9a-f]{4}-10-9a-f1{4}-[0-9a,]{4}-[0-9a-f]{12}'
    category: token
    description: AMS Auth Token

  - definition: 'key-[0-9a-zA-Z]{32}'
    category: token
    description: Mailgun API key

  - definition: '[0-9a-f]{32}-us[0-9]{1,2}'
    category: token
    description: MailChimp API key
```
These are sets of distinct rules that could be added to improve default leak discovery.

Signed-off-by: Quoc Trung Hoang <quoc-trung.hoang@etu.utc.fr>

### DIFF
--- a/ui/backend/rules.yml
+++ b/ui/backend/rules.yml
@@ -1,41 +1,49 @@
+apiVersion: v1
 rules:
-
-  - regex: -----BEGIN RSA PRIVATE KEY-----[\r\n]+(?:\w+:.+)*[\s]*(?:[0-9a-zA-Z+\/=]{64,76}[\r\n]+)+[0-9a-zA-Z+\/=]+[\r\n]+-----END RSA PRIVATE KEY----
+  - definition: -----BEGIN [A-Z]+ PRIVATE KEY( BLOCK)?-----
     category: crypto_key
-    description: rsa private key
+    description: private key
 
-  - regex: -----BEGIN EC PRIVATE KEY-----[\r\n]+(?:\w+:.+)*[\s]*(?:[0-9a-zA-Z+\/=]{64,76}[\r\n]+)+[0-9a-zA-Z+\/=]+[\r\n]+-----END EC PRIVATE KEY----
-    category: crypto_key
-    description: ec private key
-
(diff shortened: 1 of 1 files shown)


marcorosa (changes requested):
Hi,

First, we need to call the regex `regex` and not `definition` otherwise we need also to change the client.

What is `apiVersion: v1`?

_private keys_
I agree, the all-in-one rule that you wrote is better

_passwords_
It's ok having those keywords matching a lot of false positive passwords. We are aware of it, that's why we have the ML models filtering out the false positive. We know there are a lot of cases where we match irrelevant code (for example `passed`, or `pass` that is very common in python). The Snippet Model should resolve most of them.
This is the added value of Credential Digger: without passwords and ML models we become yet another GitHub token scanner (performing worse since we haven't a token entropy checker yet). 

Please reintroduce the passwords and change the `regex` key.

ichbinfrog:
I reintroduced the `pass` regex and changed the `regex` key.
The `apiVersion: v1` introduced was a mistake although I think it could be useful to version the rule files, it would allow for any future changes to the yaml structure to be backward compatible.

marcorosa (approved):

PR #48, opened 2020-09-17
Easier getting-started experience

alaabenfatma:
### Overview
This readme update provides the reader with an easier getting-started/install experience. 
It is preferred to prioritize the `pip install` approach for it being the most straightforward way of installing the tool. In other words, it is now _faster_ to _plug-and-play_ the credentialdigger and get directly to the scanning point without the need to dig deeper into the docs nor to set up a Postgres instance.
It also helps those who want to get a working instance of the _`digger`_ simply by doing CTRL-C & CTRL-V without further _`digging`_ :-)

On a side note, the `rules` wiki page has been updated.


marcorosa (approved):
I agree

PR #66, opened 2020-12-07
Merge forks' develop into internal repo

fabiosangregorio:
This PR merges changes from the fork into the new-UI branch in order to start development.
This includes:
* Possibility to load credentialdigger from local repo instead of pip
* Add root path to flask app configuration

### DIFF
--- a/ui/server.py
+++ b/ui/server.py
@@ -1,15 +1,19 @@
 import os
+import sys
 from collections import defaultdict
 
 import yaml
-from credentialdigger import PgClient, SqliteClient
 from dotenv import load_dotenv
 from flask import Flask, redirect, render_template, request, send_file
 from werkzeug.utils import secure_filename
 
-load_dotenv()
-
 APP_ROOT = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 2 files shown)


ui/server.py
```diff
@@ -176,9 +183,9 @@ def download_rule():
             'description': rule['description']
         })
 
-    with open('./backend/Downloadrules.yml', 'w') as file:
+    with open(os.path.join(APP_ROOT, './backend/Downloadrules.yml'), 'w') as file:
```
marcorosa, line 186: is this correct, or should it be changed as follows?
```
with open(os.path.join(APP_ROOT, 'backend/Downloadrules.yml'), 'w') as file:
```

fabiosangregorio: From my experience, `os.path.join` accepts a path-like object on both the arguments, and takes care of the current directory notation. I tested this both on Windows and Ubuntu.
Shall I change this and remove the `./` just to be sure?

marcorosa: It's fine, let's keep it

PR #67, opened 2020-12-08
Add dotenv loading when calling package as module

fabiosangregorio:
This PR loads the env variables in `.env` file when calling credentialdigger as a module (i.e. from the CLI with `python -m credentialdigger`), allowing the user to store configuration variables in the `.env` file instead of loading them manually as environment variables.


marcorosa (commented):
`load_dotenv` should load a `.env` file from the current directory. Does this PR make the CLI  only usable from credentialdigger folder, or does it still work from any path? Or does it work from any folder, but only with the db specified in the `.env`?

fabiosangregorio:
Since `load_dotenv` loads variables from `.env` file in the current directory, it can be used from any directory and everything works as before, with the only addition that if there's a `.env` file in the current directory, variables are picked from there.

I tested this by:
1. Creating a new directory (not in the credentialdigger repo directory)
    ```
    mkdir test-cd-dotenv
    cd test-cd-dotenv
    ```
1. Creating a new virtual environment
    ```
    virtualenv  -p `which python3.8` venv3.8 
    source ./venv3.8/bin/activate   
    ```
1. Installing credentialdigger from filesystem (on the PR branch)
    ```
    pip install /home/fabio/projects/credential-digger/
    ```
1. Scanning a repository (without `-sqlite`, without having a `.env` file in the directory)
    ```bash
    python -m credentialdigger scan https://github.com/sap/credential-digger
    Traceback (most recent call last):
      File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
        return _run_code(code, main_globals, None,
      File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
        exec(code, run_globals)
      File "/home/fabio/projects/test-cd-dotenv/venv3.8/lib/python3.8/site-packages/credentialdigger/__main__.py", line 21, in <module>
        plac.call(commands[command], sys.argv[1:])
      File "/home/fabio/projects/test-cd-dotenv/venv3.8/lib/python3.8/site-packages/plac_core.py", line 417, in call
        cmd, result = parser.consume(arglist)
      File "/home/fabio/projects/test-cd-dotenv/venv3.8/lib/python3.8/site-packages/plac_core.py", line 276, in consume
        return cmd, self.func(*(args + varargs + extraopts), **kwargs)
      File "/home/fabio/projects/test-cd-dotenv/venv3.8/lib/python3.8/site-packages/credentialdigger/scan.py", line 150, in scan
        dbport=int(os.getenv('DBPORT')))
    TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
    ```
1. Adding a `.env` file to the directory (and filling it out)
1. Retrying the scan
    ```bash
    python -m credentialdigger scan https://github.com/sap/credential-digger
    INFO:credentialdigger.scan:Database in use: Postgres
    INFO:credentialdigger.client:Detected 6061 discoveries.
    ```

marcorosa:
I think that, in order to avoid the exception raised in your point 4) (which could not be that easy to interpret without reading the documentation), what we could do is: allow users who want to use Postgres to also define the path of their env file. The default behaviour when this path is not defined is to search `.env` in the current folder. If no environment file is set, then we can raise an error (possibly an error like "no configuration file specified for your Postgres database" instead of the stracktrace).

```
python -m credentialdigger -conf=path/to/my-env REPO_URL -> OK
python -m credentialdigger REPO_URL -> search .env in the working directory (as in your previous proposal)
```

What do you think?

fabiosangregorio:
I totally agree on the readability of the error message. In your idea, would you notify the user (with a `logging.error`) in the `scan.py and add_rules.py`, or in the `__init__` of `client_postgres.py`? 
The former means that the error would only be visible when using credentialdigger from the CLI, the latter means that the user would also be notified when he forgets to `load_dotenv()` when using credentialdigger programmatically. 
Comparing this to `client_sqlite.py`, here the database is initialized in the `__init__` if it does not exist, so in postgres I would notify the error in `client_postgres.py`.

Regarding the conf argument, configurability out of the box is always great, so I agree with having it. We need to also maintain the possibility to have the current behaviour, though. What I mean is that as of now the user can run credentialdigger from CLI without having a `.env` file (e.g. by setting environment variables by hand), so I would not error out if a `.env` file is not found. Instead, as mentioned above, I would notify the user if the connection to postgres fails due to missing environment variables.
Following this approach we can have: 
* `.env` file in any directory (included cwd)
* environment variables loaded by hand (or from other ways such as `.bashrc`)

The only thing I'm not sure about is how to implement the `--conf` argument. Since the `.env` file does not only contain variables for postgres but also for the download path of the ML models, this means that env variables need to be loaded in all CLI commands (scan, add_rules, download, and other commands such as scan_user and scan_wiki if we're going to implement them), so the right place to place the `load_dotenv()` would be in the `__main__.py`, but I'm not sure if we can/don't know how to parse arguments there: can we use `argparse` there as well?

fabiosangregorio:
Following our telephone discussion, I implemented the new CLI unified parser using `argparse`, as well as removing the `int()` cast of the port env variable in order to get more readable error messages.

credentialdigger/__main__.py
```diff
@@ -1,20 +1,104 @@
...
+    parser_scan.add_argument(
+        'repo_url', type=str,
+        help='The URL of the git repository to be scanned.')
+    parser_scan.add_argument(
+        '--category', default=None, type=str,
+        help=' If specified, scan the repo using all the rules of this \
+            category, otherwise use all the rules in the db')
+    parser_scan.add_argument(
+        '--models', default=None, nargs='+',
+        help='A list of models for the ML false positives detection.\n Cannot \
```
marcorosa, line 75: Space not needed after `\n`

credentialdigger/download.py
```diff
@@ -30,18 +30,19 @@
...
     """ Download a model and link it to the credental digger models_data
     folder.
 
     Parameters
     ----------
     model: str
         The name of the model. It must be an environment variable.
-    **pip_args
+    pip_args
         Keyword arguments for pip.
```
marcorosa, line 42: What's the type of this parameter? Moreover, it's not **kwargs anymore so the description may need to be updated

credentialdigger/scan.py
```diff
@@ -45,90 +45,14 @@
...
+def scan(args):
     """
     Scan a git repository.
 
     Parameters
     ----------
-    *pip_args
-        Keyword arguments for pip.
+    args
+        Arguments from command line parser.
```
marcorosa, line 55: missing type (tuple?)

marcorosa (changes requested):
please update also `requirements.txt` since `plac` is not used anymore

marcorosa (changes requested):
Hi @fabiosangregorio, please update `requirements.txt` removing `plac` (deprecated with your PR), then I will merge.

fabiosangregorio:
@marcorosa as per our discussion, I edited the folder structure and updated requirements.txt

marcorosa (approved):

PR #74, opened 2021-01-14
Add background scanning to server and scan status to UI

fabiosangregorio:
This PR brings background scanning of repositories to the new UI, and adds live status checking of scanning.

@marcorosa if you could take a look at the threads creation and status checking in `server.py` I would be thankful.


ui/server.py
```diff
@@ -2,6 +2,7 @@
 import sys
 from collections import defaultdict
 from itertools import groupby
+import threading
```
marcorosa, line 5: fix import order

marcorosa (approved):
Thanks for this contribution!
As a suggestion, prefer snake_case naming of variables instead of CamelCase when coding in Python ;)

fabiosangregorio:
I usually follow snake case for python. The camel case variables were already present in the code and changing them by myself felt a little arbitrary. I’d be happy to convert them to snake case if you’d like me to do it :)

PR #77, opened 2021-01-18
Use last scan timestamp instead of last commit as a starting point for the scan

fabiosangregorio:
This PR deprecates the use of the last commit to start the scan and proposes the use of the timestamp of the last scan instead.
The inner workings of the `git_scanner` and `client` remain the same.
Note: to do this, the column `last_commit` in the repos table of the database gets renamed in `last_scan`.

### DIFF
--- a/credentialdigger/scanners/git_scanner.py
+++ b/credentialdigger/scanners/git_scanner.py
@@ -1,11 +1,11 @@
 import hashlib
-import shutil
 import re
+import shutil
 from datetime import datetime, timezone
 
 import hyperscan
-from git import Repo as GitRepo
 from git import NULL_TREE
+from git import Repo as GitRepo
 
 from .base_scanner import BaseScanner
 
@@ -72,9 +72,7 @@ class GitScanner(BaseScanner):
         project_path = self.clone_git_repo(git_url)
         repo = GitRepo(project_path)
 
-        try:
-            repo.rev_parse('HEAD').hexsha
-        except BadName:
+        if len(repo.branches) == 0:
             # The repository is empty
             return None, []
 
--- a/credentialdigger/scanners/git_scanner.py
+++ b/credentialdigger/scanners/git_scanner.py
@@ -63,8 +63,8 @@ class GitScanner(BaseScanner):
 
         Returns
         -------
-        str
-            The latest commit (`None` if the repository is empty)
+        int
+            The latest scan timestamp (now) (`None` if the repository is empty)
         list
             A list of discoveries (dictionaries). If there are no discoveries
             return an empty list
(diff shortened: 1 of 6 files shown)


credentialdigger/scanners/git_scanner.py
```diff
@@ -1,6 +1,7 @@
 import hashlib
 import shutil
```
marcorosa, line 2: order

credentialdigger/scanners/git_scanner.py
```diff
@@ -72,7 +73,7 @@ def scan(self, git_url, since_commit=None, max_depth=1000000):
         repo = GitRepo(project_path)
 
         try:
```
marcorosa, line 75: is this try-except block still needed to know if a repo is empty?
Can we use the `branches` variable such that we don't add useless operations?

ui/server.py
```diff
@@ -21,7 +21,7 @@
 app.config['UPLOAD_FOLDER'] = os.path.join(APP_ROOT, './backend')
 app.config['DEBUG'] = True  # Remove this line in production
 
-if os.getenv('USE_PG'):
+if os.getenv('USE_PG') is True:
```
marcorosa, line 24: is `is True` needed?

marcorosa (changes requested):

ui/server.py
fabiosangregorio: It is needed because `if os.getenv('USE_PG'):` only tests for the existence of the variable (if you have it set to `False` the condition is still true)

credentialdigger/scanners/git_scanner.py
```diff
@@ -48,32 +49,30 @@ def stream(self, rules):
...
+            The oldest timestamp to scan
         max_depth: int, optional
             The maximum number of commits to scan
 
         Returns
         -------
-        str
-            The latest commit (`None` if the repository is empty)
+        int
+            The latest scan timestamp (now) (`None` if the repository is empty)
```
marcorosa, line 67: `0` if the repository is empty (that None comes from the old method)

marcorosa (changes requested):

credentialdigger/scanners/git_scanner.py
fabiosangregorio: The method still returns `None` if the repository is empty, as you can see here:
https://github.com/SAP/credential-digger/blob/ba4d7b7b61c176e6e8b7d9187f398a3e06c1a10a/credentialdigger/scanners/git_scanner.py#L75-L77

Should it return 0 instead? In that case we wouldn't be able to know if the repo has 0 discoveries or is simply empty, but I don't think we need to differentiate the two cases.

marcorosa: No, you're right, let's leave it like it is. I misunderstood the code

PR #78, opened 2021-01-22
Add UI clients

fabiosangregorio:
In this PR `client.py`, `client_postgres.py`, and `client_sqlite.py` get extended to add more functionality to the UI, without affecting the main package code.
This is achieved by creating three new classes (`client_ui.py`, `client_ui_postgres.py`, and `client_ui_sqlite.py` in the `ui/` folder), which inherit from the main client classes.

As a use case for the need for separate classes in the main package and the UI, this PR also adds full pagination for the discoveries view, which brings down the loading time of the webpage by several seconds (based on the size of the repository).

A query for the total number of discoveries has also been created, in order to optimize loading times on the home page (using a SQL `COUNT` clause instead of selecting all discoveries and counting them from python).

### DIFF
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -334,7 +334,7 @@ class Client(Interface):
         repo_url: str
             The url of the repository
         file_name: str, optional
-            The filename to filter discoveries on
+            The name of the file to filter discoveries on
 
         Returns
         -------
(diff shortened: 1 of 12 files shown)


marcorosa (approved):
LGTM

PR #81, opened 2021-01-25
New version of the UI

fabiosangregorio:
This PR adds the new version of the UI, including:
* Revamped discoveries listing
* A new way to see discoveries: files view
* Renewed repositories view, including sorting by last scan
* Background scanning of new repositories and live status check
* Possibility to scan for private repositories with the Git token
* Server-side pagination of discoveries
* Refactoring of most HTML, javascript and css
* Refactoring of backend


credentialdigger/client.py
```diff
@@ -254,22 +250,18 @@ def get_repo(self, query, repo_url):
...
-        except self.Error:
-            self.db.rollback()
+        cursor.execute(query, (repo_url,))
+        result = cursor.fetchone()
+        if result:
+            return dict(Repo(*result)._asdict())
+        else:
             return {}
 
     def get_rules(self, category_query, category=None):
```
marcorosa, line 267: why is `category_query` positional and not optional?

credentialdigger/client.py
```diff
@@ -280,6 +272,8 @@ def get_rules(self, category_query, category=None):
         (e.g., `auth/password`). Encoding such categories in the url would
         cause an error on the server side.
 
+        NOTE: Here exceptions are suppressed in order to not stop the scanning.
+
         Parameters
```
marcorosa, line 277: missing `category_query`

marcorosa (changes requested):
I have a simple comment on the parameters of a function.

Apart from this I have some concern on the `rule_id` on delete set null: did you test this edge case? I mean: delete a rule from the db and then verify that you don't break anything else

fabiosangregorio:
Regarding the concern on `rule_id`: currently, in the main branch, there is a bug where deleting a rule does not work if a discovery generated by such rule exists in the database. This is due to the fact that we have an `ON DELETE NO ACTION` in the database. This in turn generates an exception on `/delete_rule` (because the rule id is referenced in the discoveries table) which gets suppressed by a `try-catch`.
By setting null on delete, in the UI this gets caught so that we can display the indication `(rule deleted)` in the category column of the discoveries table.

credentialdigger/client.py
fabiosangregorio: Uhm I didn't touch that, it's the same in the main branch as well, but I'll go ahead and make it optional because it has no purpose in being mandatory.

https://github.com/SAP/credential-digger/blob/dd58f5eaa6f316f50589218c7a6273555eeb18ab/credentialdigger/client.py#L270-L290

marcorosa (approved):

PR #84, opened 2021-01-27
Fix git token handling in UI

fabiosangregorio:
This PR handles the situation where the git token used for the scan is not valid for the chosen repository (either it is blank or does not the access rights).

This is handled by the `check_git_token` function in the `client_ui.py`, as I found it to be the less painful way, given that the code that tries to clone the repository is in a thread and therefore cannot communicate with the main thread easily (i.e. the `/scan_repo` POST request).

On the UI side, when the user clicks on the `Scan repo` button in the scan form modal (and therefore submits the form), a loader will appear in the button and the check will start. Once the check is finished, the loader is removed and:
* the modal hides if the check was successful
* an error will appear below the input field otherwise

The check is made with an `ls-remote` command on the repo, which I found to be the quickest way to check for the repo connection without actually cloning it.

### DIFF
--- a/ui/backend/client_ui.py
+++ b/ui/backend/client_ui.py
@@ -57,7 +57,7 @@ class UiClient(Client):
             result = cursor.fetchone()
         return files
 
-    def check_git_token(self, repo_url, git_token):
+    def check_connection(self, repo_url, git_token=None):
         """
         Check git token validity for the repository
 
@@ -65,7 +65,7 @@ class UiClient(Client):
         ----------
         repo_url: str
             The url of the repository
-        git_token: str
+        git_token: str, optional
             Git personal access token to authenticate to the git server
 
         Returns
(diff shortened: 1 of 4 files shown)


marcorosa (changes requested):
There's one major question on this approach: do you consider null tokens?
A user may not define a token for a scan, but the `check_git_token` function is called before every scan.
Thus, I assume that either the token becomes mandatory, otherwise this check will fail.

...unless we update that check like this:

```
if token and not c.check_git_token(...):
    return f'...', 401
...
```

Am I right?

fabiosangregorio:
Based on my tests, an empty token is treated by git as if it weren't there, so everything works as expected even if `git_token` is `None`. Either way, I agree on the solution seeming counterintuitive, so I went ahead and changed the method name to `check_connection`, and made the `git_token` optional.

marcorosa (approved):

PR #95, opened 2021-03-30
Tests

fabiosangregorio:
This PR introduces unit, integration, and functional testing to the package, as well as some  refactorings and several bug fixes.

## Testing

Testing has been introduced for the `GitScanner`, `FileScanner`, `Client`, `PgClient`, and `SqliteClient`. 

All the tests are contained in the `tests/` directory, with the different types of tests in the relative folder.

Tests can be run either with **unittest** or **pytest**.

```bash
python -m unittest
# or
pytest
```

Tests can be also run individually through the command:

```bash
# run all tests from a directory
pytest tests/unit_tests

# run all tests from a file
pytest tests/unit_tests/test_scans.py

# run one specific test case
pytest tests/unit_tests/test_scans.py -k test_analyze_discoveries
```

### Unit tests

Unit tests are self-contained tests that test the functionality of one single method or module, isolating it from dependencies. These tests are really fast to run and can potentially be started after every file save. They involve mocking the dependencies through `unittest.mock`

### Integration tests

Integration tests test the interaction between some of the functions or modules of the package. They are fairly quick to run, usually taking a couple of seconds to run. They can be run at every commit. It requires mocking some dependencies, and using some real-world data.

### Functional tests

Functional tests test a slice of the application, from start to finish, without mocking. These are the most powerful tests to check that the app is running correctly in the big picture, as it mirrors real-world behavior. For this, they are quite slow to run, taking over 10 seconds to run. They can be run before merging a PR to check that everything is working successfully.
Functional tests for postgres have the assumption that an initialized test database is already present, with its access stored in a `.env` file in the `tests/` folder.

## Refactorings

Some refactorings have been carried out in order to render the code testable and more readable. No API has changed to accommodate for these changes.

- **Usage of the scanners in scan methods:** First, the internal instancing of the scanner in `_scan` has been taken out of the method, in favor of passing to `_scan` an already instanced scanner. For this, `_get_scan_rules` has been created and used to instantiate the scanner with required rules in the various scan methods. This completes the IoC converting the class half dependency injection into a full instance DI, rendering the method testable.
- **Refactor client:** Methods in `client.py` have been moved around to have all related methods close together.
- **Refactor discoveries analysis:** `_analyze_discoveries` has first been taken out of the `_scan` method, in order to make it testable. Then, the method has been refactored to remove duplicated code for the debug and non-debug environments.
- **Refactor scan:** The `_check_snippet_with_generator` method has been extracted from `_scan` in order to make the function shorter and more readable.
- **Make CLI main function testable:** The CLI main parser function now takes the args as a parameter instead of taking it directly from `sys`, in order to make it testable.

## Bug fixes

- **Fix bug where discoveries in latest commit would not be detected on rescans**: this was due to testing the committed date of `curr_commit` instead of `prev_commit` against the given `since_timestamp` argument.
- **Fix bug where discoveries would get duplicated in wiki scans:** `scan_wiki` would set `force=False` in the internal call to `_scan`, resulting in previous discoveries not being deleted before rescan.
- **Fix bug where the generator would run even if no non-FP discoveries remained:** a previous commit by me introduced this suboptimization, because the generation start would test for the presence of any discoveries after analysis instead of non-FP discoveries.


fabiosangregorio:
Something I forgot to mention: a `.env` file is necessary in the `functional_tests` folder in order to run PostgreSQL functional tests.

credentialdigger/client.py
```diff
@@ -988,3 +938,37 @@ def _generate_snippet_extractor(self, repo_url):
         """
         eg = ExtractorGenerator()
         return eg.generate_leak_snippets(repo_url)
+
+    def _check_snippet_with_generator(self, generate_snippet_extractor, models):
```
marcorosa, line 942: missing documentation

credentialdigger/client.py
```diff
@@ -988,3 +938,37 @@ def _generate_snippet_extractor(self, repo_url):
...
+                # If the SnippetModel is not chosen, but the generator flag is
+                # set to True, do not generate the model (to save time and
+                # resources)
+                logger.debug(
+                    'generate_snippet_extractor=True but SnippetModel '
+                    'is not in the chosen models. No extractor to generate.')
+        return False
+
+    def _get_scan_rules(self, category=None, exclude=None):
+        """ Get the rules of the `category`, filtered by `exclude` """
```
marcorosa, line 964: improve documentation of this method

tests/functional_tests/test_scans_postgres.py
```diff
@@ -0,0 +1,57 @@
...
+import unittest
+
+from credentialdigger.cli import cli
+from git import Repo as GitRepo
+from psycopg2 import connect
+
+
+class TestScansPostgres(unittest.TestCase):
+    dotenv = os.path.join(os.path.dirname(os.path.abspath(__file__)), ".env")
+    repo_url = 'https://github.com/fabiosangregorio/credential-digger-tests'
```
marcorosa, line 13: we keep it here at the moment, but we will probably need to update this url. Nothing personal, just compliancy 😄

marcorosa (changes requested):
Additional comment/request: please copy the explanation/documentation in the wiki of the project such that we will not lose it

PR #96, opened 2021-04-09
[Feat/https] HTTPS Support

alaabenfatma:
### 🔐 HTTPS SUPPORT

A user can now easily run the web application over HTTPS. It is not mandatory, but it highly recommended.

#### How to
1. We assume you have a signed certificate (certificate.pem) and a private key (key.pem)
 a.  You can run this command to generate a certificate and private key: `openssl req -newkey rsa:2048 -new -nodes -x509 -days 365 -keyout key.pem -out cert.pem`
2. Set up the path of both the certificate and the key in your .env file
```bash
SSL_certificate=/credential-digger-ui/$PATH_to_certificate # i.e /credential-digger-ui/cert.pem
SSL_private_key=/credential-digger-ui/$PATH_to_key # i.e /credential-digger-ui/key.pem
```
3. Start the docker container
```bash
sudo docker-compose up --build
```
If everything works out just fine, you should get this output: `🔐 HTTPS will be used...`

If you leave the SSL variables in the .env file empty, the UI will run on HTTP as usual; resulting in the following output: `🔓 HTTPS will not be used... (HTTP only)`

### Important
Please make sure to make an exception to your self-signed certificate on the browser you are using. Having a warning message the first time you access the web page (i.e https://0.0.0.0:5000/) is a completely normal behavior because the certificate in use is not signed by a known Certificate Authority.

### Misc
Upon the acceptance of this PR, a wiki page will be set up to explain the process in more details for the users.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -3,7 +3,10 @@ POSTGRES_PASSWORD=
 POSTGRES_DB=
 DBHOST=postgres
 DBPORT=5432
+
 path_model=https://github.com/SAP/credential-digger/releases/download/PM-v1.0.1/path_model-1.0.1.tar.gz
 snippet_model=https://github.com/SAP/credential-digger/releases/download/SM-v1.0.0/snippet_model-1.0.0.tar.gz
-SSL_certificate=
-SSL_private_key=
+
+# Enable HTTPS by setting the path of both the certificate and the private key as shown below
+SSL_certificate=/credential-digger-ui/$PATH_to_certificate # i.e /credential-digger-ui/cert.pem
+SSL_private_key=/credential-digger-ui/$PATH_to_key # i.e /credential-digger-ui/key.pem
--- a/.env.sample
+++ b/.env.sample
@@ -3,10 +3,8 @@ POSTGRES_PASSWORD=
 POSTGRES_DB=
 DBHOST=postgres
 DBPORT=5432
-
 path_model=https://github.com/SAP/credential-digger/releases/download/PM-v1.0.1/path_model-1.0.1.tar.gz
 snippet_model=https://github.com/SAP/credential-digger/releases/download/SM-v1.0.0/snippet_model-1.0.0.tar.gz
-
 # Enable HTTPS by setting the path of both the certificate and the private key as shown below
-SSL_certificate=/credential-digger-ui/$PATH_to_certificate # i.e /credential-digger-ui/cert.pem
-SSL_private_key=/credential-digger-ui/$PATH_to_key # i.e /credential-digger-ui/key.pem
+SSL_certificate=PATH_to_certificate.pem
+SSL_private_key=PATH_to_key.pem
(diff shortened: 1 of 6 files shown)


.env.sample
```diff
@@ -3,5 +3,10 @@ POSTGRES_PASSWORD=
 POSTGRES_DB=
 DBHOST=postgres
 DBPORT=5432
+
 path_model=https://github.com/SAP/credential-digger/releases/download/PM-v1.0.1/path_model-1.0.1.tar.gz
 snippet_model=https://github.com/SAP/credential-digger/releases/download/SM-v1.0.0/snippet_model-1.0.0.tar.gz
+
+# Enable HTTPS by setting the path of both the certificate and the private key as shown below
+SSL_certificate=/credential-digger-ui/$PATH_to_certificate # i.e /credential-digger-ui/cert.pem
```
marcorosa, line 11: `/credential-digger-ui/cert.pem` are not actual folders of user's file system. This path is the path that will be created in the docker container, but a standard user doesn't have visibility over this. So, I suggest to leave the path here as the path where to find the certificate (and key) on the local file system, and then they will be put in a suitable folder in the docker container.
In practice, as it happens with the `rules.yaml` file for the rules

ui/run.sh
```diff
@@ -16,5 +16,24 @@ else
     python -m credentialdigger download snippet_model
 fi
 
+# Set HTTPS flag to True
+HTTPS=true
+
+if [[ -z "${SSL_certificate}" ]]; then
```
marcorosa, line 22: one or two [[ ? You have used one above (line 5) and two here

ui/run.sh
```diff
@@ -16,5 +16,24 @@ else
...
+    # No certificate
+    HTTPS=false
+fi
+
+if [[ -z "${SSL_private_key}" ]]; then
```
marcorosa, line 27: same comment

ui/server.py
```diff
@@ -328,4 +328,5 @@ def update_discovery_group():
         return 'OK', 200
 
 
-app.run(host='0.0.0.0', port=5000)
+if __name__ == "__main__":
```
marcorosa, line 331: single quotes for strings in python (in fact you have single quotes at the following line). Double quotes are used for docstrings

ui/wsgi.py
```diff
@@ -0,0 +1,4 @@
+from server import app
+
+if __name__ == "__main__":
```
marcorosa, line 3: here again please use single quotes :D

marcorosa (changes requested):
Hi, thanks for the PR.
There are some parts to fix, but overall it's good. 

PS: You can easier your life with a linter

marcorosa (changes requested):
The file `run.sh` still has mixed cases of [[ and [ in the if statements. Please make them uniform, then I'll merge.

NB for the path to the cert/key: is this path absolute or relative? Or can it be any of them? When you write the wiki page, please specify this :)

alaabenfatma:
> The file `run.sh` still has mixed cases of [[ and [ in the if statements. Please make them uniform, then I'll merge.
> 
> NB for the path to the cert/key: is this path absolute or relative? Or can it be any of them? When you write the wiki page, please specify this :)

You are right, missed the single brackets, sorry. 
The cert/key path is relative to the wsgi.py file, I will mention this in the wiki page.

marcorosa (approved):

marcorosa:
There's a broken test. I will fix the reference with another PR and then will merge also this one

marcorosa:
NB: verify tests after merge

Issue #94, opened 2021-03-26
Secure UI connections


marcorosa:
Add HTTPS support for the UI

Furthermore, the UI is now accessible by everybody who have visibility over a certain IP. We may secure this with a simple *optional* login page (e.g., `ui_password` set in `.env` to be entered).

marcorosa:
Fixed in #103

PR #100, opened 2021-04-20
Feat/login: Secure UI connection

alaabenfatma:
Targeted issue:
#94 
Builds upon:
#96 

### Feats
This PR introduces a new secure UI for accessing the tool. 
When the UI starts while HTTPS support is enabled, the UI will force the user to login in order to use the tool.  

### How to run
Nothing changes:  the users can just start the docker container without any changes, this will result in a login page asking them to insert an authentication key. They can use the generic password `admin` (please remember to change it).

### How to customize

- The users have to set up their custom password in`.env` by modifying the `AUTH_KEY` variable which is set to `admin` by default.
- The secret key (SECRET_KEY variable in the `.env` file) has to be generated again by the user using the command: `python -c 'import os; print(os.urandom(24).hex())'`. This SECRET_KEY will be used to securely sign the JWT (Json Web Token).

> #### ⚠️**NEVER REVEAL THE SECRET KEY NOR THE PASSWORD**

### Misc
A wiki page; detailing the process of customizing the login experience, will be written upon the merge of this PR.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -8,3 +8,5 @@ snippet_model=https://github.com/SAP/credential-digger/releases/download/SM-v1.0
 # Enable HTTPS by setting the path of both the certificate and the private key as shown below
 SSL_certificate=PATH_to_certificate.pem
 SSL_private_key=PATH_to_key.pem
+AUTH_KEY=admin # Password to be used to log in
+SECRET_KEY=# To be generated using python -c 'import os; print(os.urandom(24).hex())'
--- a/.env.sample
+++ b/.env.sample
@@ -5,7 +5,6 @@ DBHOST=postgres
 DBPORT=5432
 path_model=https://github.com/SAP/credential-digger/releases/download/PM-v1.0.1/path_model-1.0.1.tar.gz
 snippet_model=https://github.com/SAP/credential-digger/releases/download/SM-v1.0.0/snippet_model-1.0.0.tar.gz
-# Enable HTTPS by setting the path of both the certificate and the private key as shown below
 SSL_certificate=PATH_to_certificate.pem
 SSL_private_key=PATH_to_key.pem
 AUTH_KEY=admin
--- a/.env.sample
+++ b/.env.sample
@@ -8,5 +8,5 @@ snippet_model=https://github.com/SAP/credential-digger/releases/download/SM-v1.0
 # Enable HTTPS by setting the path of both the certificate and the private key as shown below
 SSL_certificate=PATH_to_certificate.pem
 SSL_private_key=PATH_to_key.pem
-AUTH_KEY=
+AUTH_KEY=admin
 SECRET_KEY=
--- a/.env.sample
+++ b/.env.sample
@@ -8,5 +8,5 @@ snippet_model=https://github.com/SAP/credential-digger/releases/download/SM-v1.0
 # Enable HTTPS by setting the path of both the certificate and the private key as shown below
 SSL_certificate=PATH_to_certificate.pem
 SSL_private_key=PATH_to_key.pem
-AUTH_KEY=admin # Password to be used to log in
-SECRET_KEY=# To be generated using python -c 'import os; print(os.urandom(24).hex())'
+AUTH_KEY=
+SECRET_KEY=
--- a/ui/server.py
+++ b/ui/server.py
@@ -28,8 +28,7 @@ app.config['DEBUG'] = True  # Remove this line in production
 app.config['SECRET_KEY'] = os.getenv('SECRET_KEY')
 
 # HTTPS = True if both a certificate and private key exist, False otherwise.
-HTTPS = os.getenv('SSL_certificate') != '' and os.getenv(
-    'SSL_private_key') != ''
+HTTPS = os.getenv('SSL_certificate') and os.getenv('SSL_private_key')
 
 if os.getenv('USE_PG') == 'True':
     app.logger.info('Use Postgres Client')
--- a/ui/server.py
+++ b/ui/server.py
@@ -25,6 +25,11 @@ app = Flask('__name__', static_folder=os.path.join(APP_ROOT, './res'),
             template_folder=os.path.join(APP_ROOT, './templates'))
 app.config['UPLOAD_FOLDER'] = os.path.join(APP_ROOT, './backend')
 app.config['DEBUG'] = True  # Remove this line in production
+app.config['SECRET_KEY'] = os.getenv("SECRET_KEY")
+
+# HTTPS = True if both a certificate and private key exist, False otherwise.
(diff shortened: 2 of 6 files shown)


.env.sample
```diff
@@ -7,4 +7,6 @@ path_model=https://github.com/SAP/credential-digger/releases/download/PM-v1.0.1/
 snippet_model=https://github.com/SAP/credential-digger/releases/download/SM-v1.0.0/snippet_model-1.0.0.tar.gz
 # Enable HTTPS by setting the path of both the certificate and the private key as shown below
 SSL_certificate=PATH_to_certificate.pem
-SSL_private_key=PATH_to_key.pem
\ No newline at end of file
+SSL_private_key=PATH_to_key.pem
+AUTH_KEY=admin # Password to be used to log in
```
marcorosa, line 11: Don't put comments here to avoid that someone leaves them as passwords

.env.sample
```diff
@@ -7,4 +7,6 @@ path_model=https://github.com/SAP/credential-digger/releases/download/PM-v1.0.1/
...
+SECRET_KEY=# To be generated using python -c 'import os; print(os.urandom(24).hex())'
```
marcorosa, line 12: same comment as above

.env.sample
marcorosa: it's better to explain these 2 points in the documentation

ui/server.py
```diff
@@ -61,8 +67,69 @@ def _get_rules():
...
+        redirect(url_for('root'))
+    else:
+        if(request.method == 'POST'):
+            auth_key = request.form['auth_key']
+            if(auth_key != os.getenv('AUTH_KEY')):
+                redirect(url_for('login'))
+                return render_template('login.html',
+                                       msg='❌ Wrong key, please try again:')
+            # We generate a UUID to be saved as a JWT's value
+            import uuid
```
marcorosa, line 103: why do you import it here? I think you can put it with the others at the first lines of the file

ui/server.py
```diff
@@ -61,8 +67,69 @@ def _get_rules():
...
 
+# Store JWT's value for every connected user
+registered_tokens = []
+
+
+@app.before_request
+def before_request():
+    """
+        Treat all incoming requests before-hand. 
+        If the user is not yet logged in, he/she will be redirected towards the login page.
```
marcorosa, line 78: No indentation here

ui/server.py
```diff
@@ -23,6 +24,11 @@
             template_folder=os.path.join(APP_ROOT, './templates'))
 app.config['UPLOAD_FOLDER'] = os.path.join(APP_ROOT, './backend')
 app.config['DEBUG'] = True  # Remove this line in production
+app.config['SECRET_KEY'] = os.getenv("SECRET_KEY")
+
+# HTTPS = True if both a certificate and private key exist, False otherwise.
+HTTPS = (os.getenv("SSL_certificate") !=
```
marcorosa, line 30: use single quotes 😛

ui/server.py
```diff
@@ -61,8 +67,69 @@ def _get_rules():
...
+@app.before_request
+def before_request():
+    """
+        Treat all incoming requests before-hand. 
+        If the user is not yet logged in, he/she will be redirected towards the login page.
+    """
+    if(HTTPS):
+        token = request.cookies.get('AUTH')
+        if(token not in registered_tokens):
+            if(request.endpoint != 'login' and '/res/' not in request.path):
```
marcorosa, line 83: why is `/res/` needed in the request path?

marcorosa (changes requested):

ui/server.py
alaabenfatma: Thank you for the other comments.
Testing whether  `/res/` is part of the request.path or not is important here; there reason behind this is that the @app.before_request decorator will block access to any route besides /login; hence blocking access to these paths:

- https://0.0.0.0:5000/res/favicon/*
- https://0.0.0.0:5000/res/css/fonts.css
- https://0.0.0.0:5000/res/css/rules.css
- https://0.0.0.0:5000/res/*

Resulting in a pale page that has no CSS nor JS. That said, we make an exception to any path containing `/res/` .

ui/server.py
```diff
@@ -61,8 +68,68 @@ def _get_rules():
...
+@app.before_request
+def before_request():
+    """
+    Treat all incoming requests before-hand. 
+    If the user is not yet logged in, he/she will be redirected towards the login page.
+    """
+    if(HTTPS):
+        token = request.cookies.get('AUTH')
+        if(token not in registered_tokens):
+            if(request.endpoint != 'login' and '/res/' not in request.path):
```
marcorosa, line 84: again check use of parentheses

ui/server.py
```diff
@@ -23,6 +25,11 @@
             template_folder=os.path.join(APP_ROOT, './templates'))
 app.config['UPLOAD_FOLDER'] = os.path.join(APP_ROOT, './backend')
 app.config['DEBUG'] = True  # Remove this line in production
+app.config['SECRET_KEY'] = os.getenv('SECRET_KEY')
+
+# HTTPS = True if both a certificate and private key exist, False otherwise.
+HTTPS = (os.getenv('SSL_certificate') !=
+         '' and os.getenv('SSL_private_key') != '')
```
marcorosa, line 32: In python, parentheses are not needed in if-clauses (unless you need more than 1 line) - pep8
Moreover, `if os.getenv('SSL_certificate') and os.getenv('SSL_private_key'):` should be sufficient

ui/server.py
```diff
@@ -61,8 +68,68 @@ def _get_rules():
...
+@app.route('/login', methods=['POST', 'GET'])
+def login():
+    # If the HTTPS protocol is not in use, then the login feature will be disabled.
+    if(not HTTPS):
+        redirect(url_for('root'))
+    else:
+        if(request.method == 'POST'):
+            auth_key = request.form['auth_key']
+            if(auth_key != os.getenv('AUTH_KEY')):
+                redirect(url_for('login'))
```
marcorosa, line 100: same comment on the if-clauses

ui/server.py
```diff
@@ -61,8 +68,68 @@ def _get_rules():
...
+                return render_template('login.html',
+                                       msg='🔒 Enter your secret key to access the scanner:')
+
 # ################### ROUTES ####################
 
+
+@app.route('/login', methods=['POST', 'GET'])
+def login():
+    # If the HTTPS protocol is not in use, then the login feature will be disabled.
+    if(not HTTPS):
```
marcorosa, line 94: `if not HTTPS`

marcorosa (changes requested):
I have some last small changes required (pep8 style rules). Then I'll merge

marcorosa (approved):

PR #107, opened 2021-05-20
Feat/similarity

melisandequentzel:
Add snippet_similarity.py + add update_similar_snippets method to client_sqlite.py + update requirements.txt

### DIFF
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -12,7 +12,6 @@ from .generator import ExtractorGenerator
 from .models.model_manager import ModelManager
 from .scanners.file_scanner import FileScanner
 from .scanners.git_scanner import GitScanner
-import .snippet_similarity.py
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.INFO)
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -1015,9 +1015,9 @@ class Client(Interface):
                                 repo_url,
                                 file_name=None,
                                 threshold=0.96):
-        """ Find snippets that are similar to the target 
+        """ Find snippets that are similar to the target
         snippet and update their state.
-        
+
         Parameters
         ----------
         target_snippet: str
@@ -1027,17 +1027,17 @@ class Client(Interface):
         file_name: str
             restrict to a given file the search for similar snippets
         threshold: double
-            update snippets with similarity score above threshold. 
+            update snippets with similarity score above threshold.
             Values lesser than 0.94 do not generally imply any relevant
-            amount of similarity between snippets, and should 
+            amount of similarity between snippets, and should
             therefore not be used.
-        
+
         Returns
         -------
         int
             The number of similar snippets found and updated
         """
-         
+
         discoveries = self.get_discoveries(repo_url, file_name)
         model = build_embedding_model()
         """ Compute target snippet embedding """
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -12,6 +12,7 @@ from .generator import ExtractorGenerator
 from .models.model_manager import ModelManager
 from .scanners.file_scanner import FileScanner
 from .scanners.git_scanner import GitScanner
+from .snippet_similarity import build_embedding_model,compute_similarity,compute_snippet_embedding
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.INFO)
@@ -1006,3 +1007,21 @@ class Client(Interface):
             raise ValueError('No rules found')
 
         return rules
+
+
+    def update_similar_snippets(self,target_snippet,state,repo_url,file_name=None,threshold=0.95):
+        discoveries = self.get_discoveries(repo_url, file_name)
+        model = build_embedding_model()
+        target_snippet_embedding = compute_snippet_embedding(target_snippet, model)
+        n_updated_snippets = 0
+        similar_ids = []
+        for d in discoveries:
+            if d['state'] == 'new':
+                snippet_embedding = compute_snippet_embedding([d['snippet']],model)
+                similarity = compute_similarity(target_snippet_embedding,snippet_embedding)
+                if similarity > threshold:
+                    n_updated_snippets += 1
+                    similar_ids.append(d['id'])                    
+        self.update_discoveries(similar_ids,state)
+        return n_updated_snippets
+
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -13,9 +13,9 @@ from .models.model_manager import ModelManager
 from .scanners.file_scanner import FileScanner
 from .scanners.git_scanner import GitScanner
 from .snippet_similarity import (
-                  build_embedding_model,
-                  compute_similarity,
-                  compute_snippet_embedding)
+    build_embedding_model,
+    compute_similarity,
+    compute_snippet_embedding)
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.INFO)
@@ -1020,20 +1020,20 @@ class Client(Interface):
         discoveries = self.get_discoveries(repo_url, file_name)
         model = build_embedding_model()
         target_snippet_embedding = compute_snippet_embedding(
-                                 target_snippet,
-                                 model)
+                                target_snippet,
+                                model)
         n_updated_snippets = 0
         similar_ids = []
         for d in discoveries:
             if d['state'] == 'new':
                 snippet_embedding = compute_snippet_embedding(
-                                  [d['snippet']],
-                                  model)
+                                 [d['snippet']],
+                                 model)
                 similarity = compute_similarity(
-                           target_snippet_embedding,
-                           snippet_embedding)
+                          target_snippet_embedding,
+                          snippet_embedding)
                 if similarity > threshold:
                     n_updated_snippets += 1
                     similar_ids.append(d['id'])
-        self.update_discoveries(similar_ids,state)
+        self.update_discoveries(similar_ids, state)
         return n_updated_snippets
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -182,6 +182,7 @@ class Client(Interface):
         with open(filename, 'r') as f:
             data = yaml.safe_load(f)
         for rule in data['rules']:
+            print(rule['regex'],rule['category'])
             self.add_rule(rule['regex'],
                           rule['category'],
                           rule.get('description', ''))
@@ -1016,7 +1017,7 @@ class Client(Interface):
                                 state,
                                 repo_url,
                                 file_name=None,
-                                threshold=0.95):
+                                threshold=0.96):
         discoveries = self.get_discoveries(repo_url, file_name)
         model = build_embedding_model()
         target_snippet_embedding = compute_snippet_embedding(target_snippet,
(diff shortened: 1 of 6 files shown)


marcorosa:
Hi @melisandequentzel thanks for the PR
Please change the base branch with `develop` instead of `main`

Then please follow all the indications of @pep8speaks that you see above to fix the style of the code. Then I will go into all the changes 😄

melisandequentzel:
Hi @marcorosa thanks for the remarks, will make the changes to the style shortly :)

melisandequentzel:
@marcorosa done, the last one: Line 2:1: F401 'tensorflow_text' imported but unused 
needs to remain because without that import some errors are raised

requirements.txt
```diff
@@ -14,3 +14,6 @@ requests
 scikit-learn
 srsly
 tqdm
+tensorflow==2.4.1
```
marcorosa, line 17: please add the dependencies in alphabetical order

credentialdigger/snippet_similarity.py
```diff
@@ -0,0 +1,38 @@
+import tensorflow as tf
+import tensorflow_text
+import logging
+import tensorflow_hub as hub
+import tensorflow.keras.preprocessing.text
+import numpy as np
```
marcorosa, line 6: The order of the imports is wrong (you can use [`isort`](https://pycqa.github.io/isort/) to do it automatically)

credentialdigger/snippet_similarity.py
```diff
@@ -0,0 +1,38 @@
...
+import tensorflow_text
+import logging
+import tensorflow_hub as hub
+import tensorflow.keras.preprocessing.text
+import numpy as np
+
+logging.getLogger('tensorflow').setLevel(logging.ERROR)
+
+
+def build_embedding_model():
```
marcorosa, line 11: missing documentation

credentialdigger/snippet_similarity.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def compute_snippet_embedding(snippet, model):
+    small_bert_result = tf.squeeze(model(tf.constant(snippet)))
+    small_bert_embeddings = small_bert_result.numpy()
+    snippet_embedding = np.mean(small_bert_embeddings, axis=0)
+    return snippet_embedding
+
+
+def compute_similarity(emb_1, emb_2):
```
marcorosa, line 35: same as above

credentialdigger/client.py
```diff
@@ -1006,3 +1010,25 @@ def _get_scan_rules(self, category=None, exclude=None):
             raise ValueError('No rules found')
 
         return rules
+
+    def update_similar_snippets(self,
+                                target_snippet,
+                                state,
+                                repo_url,
+                                file_name=None,
```
marcorosa, line 1018: why do you need the filename? As far as I understood, given a `target_snippet` you need to get all the "new" discoveries for a certain repo, and then compute each similarity and eventually flag other FPs. So what's the need of having the file name in this process?

ui/backend/client_ui.py
```diff
@@ -97,3 +101,26 @@ def check_repo(self, repo_url, git_token=None, local_repo=False):
...
+
+    def update_similar_snippets(self,
+                                target_snippet,
+                                state,
+                                repo_url,
+                                file_name=None,
+                                threshold=0.96):
+        discoveries = self.get_discoveries(repo_url,
+                                           file_name,
+                                           state_filter='new')[1]
```
marcorosa, line 113: why do you only keep the [1] instead of the whole list?
are you sure you don't raise an IndexError if you don't have other discoveries?

ui/backend/client_ui_sqlite.py
```diff
@@ -87,6 +91,9 @@ def get_discoveries(self, repo_url, file_name=None, state_filter=None,
         # Build outer query to get all occurrences of the paginated snippets
         query = 'SELECT * FROM discoveries WHERE repo_url=?'
         params = [repo_url]
+        params = [repo_url]
+        params = [repo_url]
+        params = [repo_url]
```
marcorosa, line 96: ?

ui/backend/client_ui_sqlite.py
```diff
@@ -165,3 +172,26 @@ def get_files_summary(self, repo_url):
                 " FROM discoveries WHERE repo_url=?"
                 " GROUP BY file_name"
             ))
+
+    def update_similar_snippets(self,
```
marcorosa, line 176: why do you need to redefine this (or to copy it from the other class)?

marcorosa:
Hi, apart from all the changes/doubts requested above, I have 2 further concerns?
- if/when the PR is ready to be reviewed, remember to turn it to "ready for review" and don't leave it draft
- There are few comments, and the code is hard to follow. Please write comments in the code, and remember to write the doctrings for ALL your new methods ;)

requirements.txt
```diff
@@ -14,3 +14,6 @@ requests
...
+tensorflow-text==2.4.1
+tf-models-official
```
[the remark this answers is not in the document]
melisandequentzel: Version 2.4.1 isn't available for tf-models-official, the right version is picked up automatically. The rule to have matching versions only applies to tf and tf-text

credentialdigger/client.py
melisandequentzel: The purpose of having a filename here is to allow for exclusive updating of similar snippets found in the same file as the target snippet, because there could be a situation where a file contains similar snippets as another file, but one with real passwords and the other with dummy ones. This parameter is provided 'just in case' someone needs this feature, but is not meant to be used as a general rule. I can remove it if you think it should not be there.

ui/backend/client_ui.py
melisandequentzel: I added [1] because the get_discoveries method from the UI classes returns the number of discoveries as first element, which is useless in this case. I'm only retriving the actual discoveries, not their count. This is what adding [1] allows for.

ui/backend/client_ui_sqlite.py
melisandequentzel: Precisely because of the get_discoveries method that is specific to the UI classes. No other way to go than to rewrite the adapted version of update_similar_snippets in the UI classes

melisandequentzel: Or maybe checking for length of method output directly in client.py, but doesn't seem any more efficient

melisandequentzel:
Hi, okay, thank you for the reviews. I will add comments + docstrings, and will make sure to do so and to have the right order for imports in the future :)

credentialdigger/client.py
marcorosa: Let's keep it with this PR then, but before merging into the main branch I would like to test it. Moreover, I'm not fully convinced this parameter is really needed: if the same snippet appears in 2 different files, then either they are treated at the same way (i.e., both TP or both FP), or it is the path model that should flag one of the two as a FP.

ui/backend/client_ui_sqlite.py
```diff
@@ -165,3 +169,26 @@ def get_files_summary(self, repo_url):
...
+                                target_snippet,
+                                state,
+                                repo_url,
+                                file_name=None,
+                                threshold=0.95):
```
marcorosa, line 178: it was 0.96 in the parent class. Why have you changed it here?

credentialdigger/client.py
```diff
@@ -1006,3 +1008,51 @@ def _get_scan_rules(self, category=None, exclude=None):
...
+            if d['state'] == 'new':
+                """ Compute snippet embedding """
+                snippet_embedding = compute_snippet_embedding(d['snippet'],
+                                                              model)
+                """ Compute similarity of target snippet and snippet """
+                similarity = compute_similarity(target_snippet_embedding,
+                                                snippet_embedding)
+                if similarity > threshold:
+                    n_updated_snippets += 1
+                    self.update_discovery(d['id'], state)
```
marcorosa, line 1057: does it work without passing the query like in the update_discovery method?

marcorosa (changes requested):
You are missing the client_ui_postgres class

credentialdigger/client.py
melisandequentzel: Yes in theory, but then why need a snippet model if the path model is always right? Some snippets with fake credentials can appear in paths that aren't dedicated to testing and/or detected as dummy by the path model, since we're applying a model to distinguish real from fake passwords after the path model. The snippet model may fail to recognize a fake password as being fake, and that's the whole point of having a similarity model: to provide a simple, 'user-powered' way of compensating for failings/shortcomings of one of the two models. This may become superfluous when the snippet model reaches a very high accuracy, but seems quite relevant as of now. Adding the filename parameter offers a more granular control over the update, which can't be a bad thing. In the UI, it could just take the form of a button to tick 'restrict updates to current file'. I noticed, there are often similar snippets in the same file, so it would still be efficient using it that way.

melisandequentzel: (There clearly is a (very) small danger bound to the similarity feature, and steps to mitigate this danger should be taken, I think. Otherwise, we'll need to clearly inform the user on the potential risks of overly using this functionality on snippets that do contain a password).

credentialdigger/client.py
```diff
@@ -1006,3 +1008,51 @@ def _get_scan_rules(self, category=None, exclude=None):
...
+        ----------
+        target_snippet: str
+        state: str
+            state to update similar snippets to
+        repo_url: str
+        file_name: str
+            restrict to a given file the search for similar snippets
+        threshold: double
+            update snippets with similarity score above threshold.
+            Values lesser than 0.94 do not generally imply any relevant
```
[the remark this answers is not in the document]
melisandequentzel: 0.94, it's intended as a guideline on which values to use for this parameter. 0.96 is usually perfect, but in some rare cases, putting a lower value will ensure all similar snippets will be detected; however that value should always be greater than 0.94.

ui/backend/client_ui.py
```diff
@@ -97,3 +101,49 @@ def check_repo(self, repo_url, git_token=None, local_repo=False):
...
+        ----------
+        target_snippet: str
+        state: str
+            state to update similar snippets to
+        repo_url: str
+        file_name: str
+            restrict to a given file the search for similar snippets
+        threshold: double
+            update snippets with similarity score above threshold.
+            Values lesser than 0.94 do not generally imply any relevant
```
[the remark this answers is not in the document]
melisandequentzel: 0.94

ui/backend/client_ui_sqlite.py
melisandequentzel: Indeed, I forgot to update the value here, I thought I had done so. It should be 0.96

credentialdigger/client.py
melisandequentzel: Yes, because the update_discovery method that gets called is that of the child class, which doesn't require a query parameter. It definitely works.

ui/backend/client_ui_sqlite.py
melisandequentzel: Oh but the code should have been removed from that class altogether, that's why I didn't update it. I'm removing it now

melisandequentzel:
@marcorosa I'm not missing the code for the client_ui_postgres file because it's in the parent class (client_ui file). I had failed to remove the code from the client_ui _sqlite file. I just did so, and re-tested it.

melisandequentzel:
@marcorosa I fixed the float types and removed code from client_ui_sqlite, so the code should be ready by now. I replied to your comment regarding the filename parameter in the update_similar_snippets method. Should I make the change anyway?

marcorosa (approved):

PR #112, opened 2021-06-07
Feat/similarity UI

melisandequentzel:


ichbinfrog:
Hi @melisandequentzel , I think it's better if you don't push for every commit (we receive a lot of mail if we're subscribed to the repo). If you want pep8 linting I suggest you use [pyblack](https://github.com/psf/black) locally.
Cheers

PR #116, opened 2021-06-10
Feat/similarity performance

melisandequentzel:


marcorosa:
Hi, I quickly checked the code changes. I see that the proposal contained in the PR is not acceptable as is, since there is a major change in the database in order to support a feature, and this will cause a breaking change for all the users only in order to add a feature, even if an important one.
So in order to bring this proposal to an acceptable level, we need to agree on a few points.

In particular, the db cannot be changed in order to update a feature or a ml model. My suggestion here, if possible, is to limit the impact to the new function/model only. An idea could be to dedicate a sqlite db for the embeddings and similarity related stuff. Otherwise we may think of dedicating a table in the db. it's hard to say now before having a clear picture in my mind.

My suggestion is to change the PR in order to provide the new similarity feature only, without focus on the performance, and later we will optimize it (and consider whether to add a table or to use a second sqlite db, or breaking change the code). Maybe what I'm asking is same as the PR #112 that I haven't had the time to review before you closed it. If so, you may reopen it, and I will review it. Then, we'll find a way not to make the embeddings an unbearable burden.

melisandequentzel:
Hi, okay I see. In my mind, at least for future users, adding a column in a database seemed less of a hassle than adding a completely new database, especially since anyway be it for similarity or something else, storing snippets embedding is likely to be necessary. But it clearly made more sense to serialize the database and have a new table with embeddings, than to add a column. Having a completely new db will likely make retrieving embeddings less straightforward (than keeping them in the same db in another table), since we'll need to make sure the ids are the same across dbs, and it may have implications on performance.
Yes what you are asking about is pretty much what's in PR#112, but I feel not storing embeddings is a bit ridiculous. I think I can manage the timeout with little work (to have the page automatically reload when it's ready), but still the performance will not only be slow (and prohibitively so on large repos), but really inefficient since the same embeddings for all snippets will be recomputed each time the method is called instead of being computed once, then stored and loaded.

marcorosa:
Ok, then since you already did all the work to both calculate and store embeddings, please dedicate a table in the db for the embeddings (without modifying the `discoveries` one). Then you can reopen the PR and I'll review it

melisandequentzel:
Okay sounds good

PR #117, opened 2021-06-14
Feat/similarity table

melisande1:


credentialdigger/client.py
```diff
@@ -453,6 +484,17 @@ def get_discovery_group(self, query, state_query, repo_url, state=None):
             cursor.execute(query, (repo_url,))
         return cursor.fetchall()
 
+    def get_embedding(self, query, discovery_id=None, snippet=None):
```
marcorosa, line 487: missing documentation

credentialdigger/client.py
```diff
@@ -453,6 +484,17 @@ def get_discovery_group(self, query, state_query, repo_url, state=None):
...
+        cursor = self.db.cursor()
+        try:
+            if discovery_id:
+                cursor.execute(query, (discovery_id,))
+            else:
```
marcorosa, line 492: What happens if both the `discovery_id` and the `snippet` are `None`?

credentialdigger/client.py
```diff
@@ -559,7 +601,7 @@ def update_discovery_group(self, query, new_state, repo_url, file_name=None,
 
     def scan(self, repo_url, category=None, models=None, exclude=None,
              force=False, debug=False, generate_snippet_extractor=False,
-             local_repo=False, git_token=None):
+             similarity=False, local_repo=False, git_token=None):
```
marcorosa, line 604: if you want to add a new parameter, then you also have to provide its documentation in the docstrings (of this method and the others where you pass it)

credentialdigger/client.py
```diff
@@ -874,6 +917,8 @@ def _scan(self, repo_url, scanner, models=None, force=False, debug=False,
             except ModuleNotFoundError:
                 logger.warning('SnippetModel not found. Skip it.')
 
+        if similarity:
```
marcorosa, line 920: if it takes some time, please add a log. Something like
`logger.info('Build embedding model for this repository')`
otherwise the scan will look like pending

credentialdigger/client.py
```diff
@@ -1017,7 +1066,6 @@ def update_similar_snippets(self,
                                 threshold=0.96):
         """ Find snippets that are similar to the target
         snippet and update their state.
-
```
marcorosa, line 1020: why have you removed the blank lines from this docstring?

credentialdigger/client.py
```diff
@@ -1031,28 +1079,29 @@ def update_similar_snippets(self,
...
             amount of similarity between snippets, and should
             therefore not be used.
-
         Returns
         -------
         int
             The number of similar snippets found and updated
         """
 
         discoveries = self.get_discoveries(repo_url, file_name)
```
marcorosa, line 1088: do you need all the discoveries or only the "new" ones?

credentialdigger/client.py
```diff
@@ -1031,28 +1079,29 @@ def update_similar_snippets(self,
...
-                snippet_embedding = compute_snippet_embedding(d['snippet'],
-                                                              model)
-                """ Compute similarity of target snippet and snippet """
-                similarity = compute_similarity(target_snippet_embedding,
-                                                snippet_embedding)
-                if similarity > threshold:
-                    n_updated_snippets += 1
-                    self.update_discovery(d['id'], state)
-        return n_updated_snippets
+        if target_embedding:
```
marcorosa, line 1092: to make this nested block more readable and easier:

```
# comment to explain it
if not target_embedding:
    return 0

# We have a target embedding (or another comment)
for d in discoveries:
    ...
```

credentialdigger/client.py
```diff
@@ -1031,28 +1079,29 @@ def update_similar_snippets(self,
...
+            for d in discoveries:
+                if (
+                    d['state'] != state
+                    and self.get_embedding(discovery_id=d['id'])
+                ):
+                    # Compute similarity of target snippet and snippet
```
marcorosa, line 1098: this comment is a little confusing: didn't we calculate the embedding of the target snippet before? 
Indeed, here you're using `target_embedding` and not `target_snippet`

credentialdigger/client.py
```diff
@@ -1031,28 +1079,29 @@ def update_similar_snippets(self,
...
+                    embedding = self.get_embedding(discovery_id=d['id'])
+                    similarity = compute_similarity(target_embedding,
+                                                    embedding)
+                    if similarity > threshold:
+                        n_updated_snippets += 1
+                        self.update_discovery(d['id'], state)
```
marcorosa, line 1104: invert these 2 lines: before you update the discovery and then you update `n`.
It doesn't change anything, but it seems more logical

credentialdigger/client_postgres.py
```diff
@@ -195,6 +198,66 @@ def add_rule(self, regex, category, description=''):
             query='INSERT INTO rules (regex, category, description) \
                     VALUES (%s, %s, %s) RETURNING id')
 
+    def add_embedding(self, discovery_id, embedding=None, repo_url=''):
+        """ Add an embedding to the embeddings table.
```
marcorosa, line 202: missing blank line before Parameters

credentialdigger/client_postgres.py
```diff
@@ -195,6 +198,66 @@ def add_rule(self, regex, category, description=''):
...
+            The id of the discovery whose embedding is being added
+        embedding: list
+            The embedding being added
+        repo_url: str
+            The discovery's repository url
+        """
+
+        cursor = self.db.cursor()
+        snippet = self.get_discovery(discovery_id)['snippet']
+        if embedding is None:
```
marcorosa, line 215: `if not embedding`:

credentialdigger/client_postgres.py
```diff
@@ -195,6 +198,66 @@ def add_rule(self, regex, category, description=''):
...
+            cursor.execute(query, (discovery_id,
+                                   embedding,
+                                   snippet,
+                                   repo_url))
+            self.db.commit()
+        except Error:
+            self.db.rollback()
+
+    def add_embeddings(self, repo_url):
+        """ Bulk add embeddings.
```
marcorosa, line 230: missing blank line

credentialdigger/client_postgres.py
```diff
@@ -195,6 +198,66 @@ def add_rule(self, regex, category, description=''):
...
+        Parameters
+        ----------
+        repo_url: str
+            The discoveries' reposiroty url
```
marcorosa, line 234: typo

credentialdigger/client_postgres.py
```diff
@@ -249,6 +313,46 @@ def delete_discoveries(self, repo_url):
             repo_url=repo_url,
             query='DELETE FROM discoveries WHERE repo_url=%s RETURNING true')
 
+    def delete_embedding(self, discovery_id):
+        """ Delete an embedding.
+        Parameters
```
marcorosa, line 318: missing blank line

credentialdigger/client_postgres.py
```diff
@@ -195,6 +198,66 @@ def add_rule(self, regex, category, description=''):
...
+        discoveries = self.get_discoveries(repo_url)
+        discoveries_ids = [d['id'] for d in discoveries]
+        snippets = [d['snippet'] for d in discoveries]
+        model = build_embedding_model()
+        embeddings = [compute_snippet_embedding(s, model) for s in snippets]
+        try:
+            query = 'INSERT INTO embeddings (id, embedding, snippet, repo_url) \
+                    VALUES (%s, %s, %s, %s);'
+            insert_tuples = []
+            for i in range(len(discoveries_ids)):
```
marcorosa, line 247: `for current_disc in discoveries:` ? Unless you need `i`

credentialdigger/client_postgres.py
```diff
@@ -249,6 +313,46 @@ def delete_discoveries(self, repo_url):
...
+            `True` if embedding was successfully deleted,
+            `False` otherwise
+        """
+        return super().delete_embedding(
+            query='DELETE FROM embeddings WHERE id=%s',
+            discovery_id=discovery_id)
+
+    def delete_embeddings(self, repo_url):
+        """ Delete all embeddings from a repository.
+        Parameters
```
marcorosa, line 335: same comment

sql/create_table.sql
```diff
@@ -28,3 +28,11 @@ CREATE TABLE discoveries (
   FOREIGN KEY (repo_url) REFERENCES repos ON DELETE CASCADE ON UPDATE CASCADE,
   FOREIGN KEY (rule_id) REFERENCES rules ON DELETE SET NULL ON UPDATE CASCADE
 );
+
+CREATE TABLE embeddings (
+  id INTEGER REFERENCES discoveries,
+  embedding FLOAT [],
+  snippet TEXT DEFAULT '',
+  repo_url TEXT,
```
marcorosa, line 36: don't they reference discoveries.snippet and repos.url?

credentialdigger/snippet_similarity.py
```diff
@@ -53,38 +53,41 @@ def compute_snippet_embedding(snippet, model):
...
 
 
 def compute_similarity(emb_1, emb_2):
     """ Compute the cosine similarity of two snippets' embeddings.
 
     Parameters
     ----------
-    emb_1: np.array
+    emb_1: list
         first snippet embedding
```
marcorosa, line 76: Initial letter should be capitalised (also for `emb_2`)

credentialdigger/snippet_similarity.py
```diff
@@ -53,38 +53,41 @@ def compute_snippet_embedding(snippet, model):
...
     small_bert_embeddings = small_bert_result.numpy()
-    """ Compute snippet's embedding as the mean of
-    the token embeddings """
+    # Compute snippet's embedding as mean of token embeddings
     snippet_embedding = np.mean(small_bert_embeddings, axis=0)
-    return snippet_embedding
+    return snippet_embedding.tolist()
 
 
 def compute_similarity(emb_1, emb_2):
```
marcorosa, line 70: please use better variables names (e.g., embedding_1 and 2), otherwise we risk not to understand it in 6 months :D

ui/res/js/discoveries.js
```diff
@@ -145,14 +145,18 @@ function initDiscoveriesDataTable() {
               ` : ""}
             </tbody>
           </table><div>`;
+          
+          const actions = discoveriesBtnGroupTemplate('Mark as')+`
```
marcorosa, line 149: actions is not a very good name. We risk to have a conflict because it's too general (and maybe prone to be used for something else)

ui/server.py
```diff
@@ -277,7 +277,8 @@ def scan_repo():
         'models': models,
         'force': force_scan,
         'git_token': git_token,
-        'local_repo': local_repo
+        'local_repo': local_repo,
+        'similarity': True
```
marcorosa, line 281: should we leave it mandatory or ask the user?

ui/server.py
```diff
@@ -407,6 +408,22 @@ def update_discovery_group():
...
+    target_snippet = request.form.get('snippet')
+    state = request.form.get('state')
+    url = request.form.get('url')
+    file = request.form.get('file')
+
+    response1 = c.update_discovery_group(state, url, file, target_snippet)
+    response2 = c.update_similar_snippets(target_snippet,
+                                          state,
+                                          url)
+    if (response1 is False or response2 is False):
```
marcorosa, line 422: ( ) not needed

ui/backend/client_ui_sqlite.py
```diff
@@ -165,3 +170,33 @@ def get_files_summary(self, repo_url):
                 " FROM discoveries WHERE repo_url=?"
                 " GROUP BY file_name"
             ))
+
+    def add_embeddings(self, repo_url):
```
marcorosa, line 174: missing documentation

ui/backend/client_ui_sqlite.py
```diff
@@ -165,3 +170,33 @@ def get_files_summary(self, repo_url):
...
+            insert_tuples = []
+            for i in range(len(discoveries_ids)):
+                insert_tuples.append((discoveries_ids[i],
+                                      snippets[i],
+                                      embedding_strings[i],
+                                      repo_url,))
+            cursor.executemany(query, insert_tuples)
+            self.db.commit()
+        except Error:
+            self.db.rollback()
```
marcorosa, line 198: what are you doing in this except block?

ui/backend/client_ui_sqlite.py
```diff
@@ -165,3 +170,33 @@ def get_files_summary(self, repo_url):
...
+        cursor = self.db.cursor()
+        discoveries = self.get_discoveries(repo_url)[1]
+        discoveries_ids = [d['id'] for d in discoveries]
+        snippets = [d['snippet'] for d in discoveries]
+        model = build_embedding_model()
+        embeddings = [compute_snippet_embedding(s, model) for s in snippets]
+        embedding_strings = ["" for i in discoveries]
+        for i in range(len(discoveries_ids)):
+            for emb in embeddings[i]:
+                embedding_strings[i] += str(emb) + ","
```
marcorosa, line 184: not very elegant. Isn't there another way to do it?

marcorosa (changes requested):
I have several comments (you can follow the review) that are "minor", but I also have a major one:
The add/get/delete embedding methods are almost the same both for the sqlite and postgres classes. So, can't we move them to the parent class like we did for other methods (e.g., add_discovery)?
In the end, the process is the same, and the only difference is in the query format.
This would considerably help the maintainability of the project on the long term.

melisande1:
Thank you for the review, will make the requested changes and will reply to specific comments.
Indeed, I will move parts of the code to the parent class in get/add/delete_embeddings, it makes more sense.

melisande1 (commented):
I replied to some comments, will make the changes accordingly (and will make changes where comments didn't require a reply)

melisande1:
@marcorosa I made all the requested changes (with the exception of adding the references in the embeddings table, need to figure out why it doesn't work)

credentialdigger/client_postgres.py
```diff
@@ -373,6 +457,37 @@ def get_discovery_group(self, repo_url, state=None):
...
             discoveries WHERE repo_url=%s GROUP BY file_name, snippet, state')
 
+    def get_embedding(self, discovery_id=None, snippet=None):
+        """ Retrieve a discovery embedding.
+
+        Parameters
+        ----------
+        discovery_id: int
+            The id of the discovery whose embedding is being retrieved
+        snippet: str
```
marcorosa, line 467: both of them are optional arguments

credentialdigger/client_postgres.py
```diff
@@ -373,6 +457,37 @@ def get_discovery_group(self, repo_url, state=None):
...
+            The snippet whose embedding is being retrieved. Only used if
+            discovery_id is not provided
+
+        Returns
+        -------
+        list
+            The embedding for the provided
```
marcorosa, line 474: a list of embeddings (floats) ?

credentialdigger/client_sqlite.py
```diff
@@ -181,6 +192,60 @@ def add_discoveries(self, discoveries, repo_url):
                 state=d['state']
             ), discoveries)
 
+    def add_embedding(self, discovery_id, embedding=None, repo_url=''):
```
marcorosa, line 195: I assume this method (and add_embeddings) can be moved (maybe partially, maybe wholly) in the parent class

credentialdigger/client_sqlite.py
```diff
@@ -398,6 +499,38 @@ def get_discovery_group(self, repo_url, state=None):
...
+            query = 'SELECT embedding FROM embeddings WHERE snippet=?'
+        else:
+            return None
+        str_embedding = super().get_embedding(query=query,
+                                              discovery_id=discovery_id,
+                                              snippet=snippet)
+        embedding = []
+        if str_embedding:
+            embedding = [float(emb)
+                         for emb in str_embedding[0].split(",")[:-1]]
```
marcorosa, line 531: didn't you use ; instead of ,?

credentialdigger/client_sqlite.py
```diff
@@ -398,6 +499,38 @@ def get_discovery_group(self, repo_url, state=None):
...
+        discovery_id: int
+            The id of the discovery whose embedding is being retrieved
+        snippet: str
+            The snippet whose embedding is being retrieved. Only used if
+            discovery_id is not provided
+
+        Returns
+        ------
+        list
+            The embedding
```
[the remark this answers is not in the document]
marcorosa: ok I understood it later

credentialdigger/client_sqlite.py
marcorosa, line 531: wasn't `str_embedding` a string?
why `str_embedding[0].split(...)`? it will split 1 letter

ui/backend/client_ui.py
```diff
@@ -115,12 +115,12 @@ def update_similar_snippets(self,
         ----------
         target_snippet: str
         state: str
-            state to update similar snippets to
+            State to update similar snippets to
         repo_url: str
```
marcorosa, line 119: missing description (even if it's trivial)

ui/backend/client_ui.py
```diff
@@ -115,12 +115,12 @@ def update_similar_snippets(self,
         ----------
         target_snippet: str
```
marcorosa, line 116: missing (trivial) description

ui/backend/client_ui.py
```diff
@@ -131,19 +131,23 @@ def update_similar_snippets(self,
             The number of similar snippets found and updated
         """
 
-        discoveries = self.get_discoveries(repo_url,
-                                           file_name,
-                                           state_filter='new')[1]
-        model = build_embedding_model()
-        target_snippet_embedding = compute_snippet_embedding(target_snippet,
-                                                             model)
+        discoveries = self.get_discoveries(repo_url, file_name)[1]
```
marcorosa, line 134: this method returns a list of dictionaries. Why are you taking [1]? if you don't have discoveries (or if you have only 1) this should raise a IndexError

ui/backend/client_ui.py
```diff
@@ -131,19 +131,23 @@ def update_similar_snippets(self,
...
+                    and self.get_embedding(discovery_id=d['id'])
+                ):
+                    # Compute similarity of target embedding and embedding
+                    embedding = self.get_embedding(discovery_id=d['id'])
+                    similarity = compute_similarity(target_embedding,
+                                                    embedding)
+                    if similarity > threshold:
+                        n_updated_snippets += 1
+                        self.update_discovery(d['id'], state)
+            return n_updated_snippets
```
marcorosa, line 153: isn't this block of code same as in the parent class?

ui/backend/client_ui_postgres.py
```diff
@@ -1,5 +1,10 @@
+from sqlite3 import Error
```
marcorosa, line 1: sqlite's Error in Postgres? Are you sure?

ui/backend/client_ui_postgres.py
```diff
@@ -162,3 +167,33 @@ def get_files_summary(self, repo_url):
                 " FROM discoveries WHERE repo_url=%s"
                 " GROUP BY file_name"
             ))
+
+    def add_embeddings(self, repo_url):
```
marcorosa, line 171: again copy-pasted code from a parent class

ui/server.py
```diff
@@ -407,6 +408,22 @@ def update_discovery_group():
...
+    target_snippet = request.form.get('snippet')
+    state = request.form.get('state')
+    url = request.form.get('url')
+    file = request.form.get('file')
+
+    response1 = c.update_discovery_group(state, url, file, target_snippet)
+    response2 = c.update_similar_snippets(target_snippet,
+                                          state,
+                                          url)
+    if response1 is False or response2 is False:
```
marcorosa, line 422: `if not response1 or not response2:`
(more clear)

ui/server.py
```diff
@@ -407,6 +408,22 @@ def update_discovery_group():
...
+        return 'Error in updating similar snippets', 500
```
marcorosa, line 423: can this happen? 
According to the `if`, you may update the similar ones without updating the "discovery group"

marcorosa: shouldn't we do something like
```
response1 = c.update_discovery_group(state, url, file, target_snippet)

if not response1:
    return '...', 500
response2 = c.update_similar_snippets(target_snippet,state,url)
if not response2:
    return '...', 500
return '...', 200
```

marcorosa (changes requested):
Hi, thanks for the wide PR!

I have some minor comments on style and structures. I think they will be easy to address.

After that, we can discuss on 2 concerns that I have.
1- At first I didn't understand why you had to use a string for embeddings in sqlite and a list of floats in postgres, then I got it. I wonder if this approach is better than using serialisation (i.e., `pickle`).
2- I'm pretty sure that we can move most/all the `add_embedding(s)` methods to Client, since they are all copy-pasted. Maybe there is some post-processing due in the children, but anyway we have to sort this out after the point at 1)

marcorosa:
By the way I don't know why @pep8speaks didn't send any message. Either you have been perfect with the style rules, or we'll receive its alert later once I'll open the PR to the main branch

melisande1:
Hi, thank you for the review! I will make all the changes. Pep did speak, its message is hidden among the commits, when you expand them you can see it: the style is fine. I have two remarks, one regarding moving the add_embeddings from the UI child classes to the regular Client: there is a difference, it's in the get_discoveries[1] vs get_discoveries, I don't think there's a way around, unless passing the discoveries as arguments to the update_discoveries method, but it will likely make things more complicated. Regarding the other tiny remark I have, I'm replying to the specific comment.

ui/server.py
melisande1: I agree, but in the update_discovery_group, there `if response is False`

ui/server.py
melisande1: yes it's better, I thought about it

ui/backend/client_ui.py
melisande1: yes this block is the same as in the parent class, but unlike the entire method, so I cannot simply delete the method from the UI class

ui/backend/client_ui_postgres.py
melisande1: not exactly copy pasted, it's the same issue as update_similar_discoveries, there's the get_discoveries[1], that I can't get rid of

credentialdigger/client.py
```diff
@@ -112,12 +113,78 @@ def add_discovery(self, query, file_name, commit_id, line_number, snippet,
...
+                                     embeddings,
+                                     [repo_url] * len(discoveries_ids)))
+            cursor.executemany(query, insert_tuples)
+            self.db.commit()
+        except self.Error:
+            self.db.rollback()
+            map(lambda disc_id, emb: self.add_embedding(disc_id,
+                                                        emb,
+                                                        repo_url=repo_url),
+                zip(discoveries_ids, embeddings))
```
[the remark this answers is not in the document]
melisande1: yes, likely, I copied the layout of add_discoveries, so I might as well, just remove the exception part 
and just do a rollback

credentialdigger/client_sqlite.py
melisande1: no, I used ,

ui/backend/client_ui.py
melisande1: No, in the UI classes, this method returns the number of discoveries as 1st element and the actual discoveries as 2nd element

credentialdigger/client_sqlite.py
melisande1: Actually there is a mistake in the docstring of get_embedding: str_embedding is a tuple containing a single string, not a string

melisande1: I will move the [0] to the parent definition, so that get_embedding does indeed return a string/list, but as is, the code works

credentialdigger/client_sqlite.py
melisande1: It's already largely in the parent class:
      return super().add_embedding(query,
                                     discovery_id,
                                     snippet,
                                     embedding_str,
                                     repo_url)

credentialdigger/client_postgres.py
```diff
@@ -373,6 +457,37 @@ def get_discovery_group(self, repo_url, state=None):
...
+            query = 'SELECT embedding FROM embeddings WHERE snippet=%s'
+        else:
+            return None
+        embedding = super().get_embedding(query=query,
+                                          discovery_id=discovery_id,
+                                          snippet=snippet)
+        if embedding:
+            return embedding[0]
+        else:
+            return None
```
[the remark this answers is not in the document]
melisande1: yes I will move it

credentialdigger/client_postgres.py
```diff
@@ -227,6 +275,7 @@ def delete_repo(self, repo_url):
         bool
             `True` if the repo was successfully deleted, `False` otherwise
         """
+        self.delete_embeddings(repo_url)
```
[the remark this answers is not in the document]
melisande1: ok

credentialdigger/client.py
```diff
@@ -1039,20 +1202,25 @@ def update_similar_snippets(self,
...
+            return 0
+        # We have a target embedding for the target snippet
+        else:
+            for d in discoveries:
+                if (
+                    d['state'] != state
+                    and self.get_embedding(discovery_id=d['id'])
+                ):
+                    # Compute similarity of target embedding and embedding
+                    embedding = self.get_embedding(discovery_id=d['id'])
```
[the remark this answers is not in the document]
melisande1: okay, I will make it one call

credentialdigger/client.py
```diff
@@ -1039,20 +1202,25 @@ def update_similar_snippets(self,
...
-                if similarity > threshold:
-                    n_updated_snippets += 1
-                    self.update_discovery(d['id'], state)
-        return n_updated_snippets
+        # If specified target snippet is not found in the embeddings table,
+        # no update is performed
+        if not target_embedding:
+            return 0
+        # We have a target embedding for the target snippet
+        else:
```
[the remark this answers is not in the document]
melisande1: okay

credentialdigger/client.py
```diff
@@ -883,11 +1042,15 @@ def _scan(self, repo_url, scanner, models=None, force=False, debug=False,
                     curr_d['file_name'], curr_d['commit_id'],
                     curr_d['line_number'], curr_d['snippet'], repo_url,
                     curr_d['rule_id'], curr_d['state'])
+                if similarity:
+                    self.add_embedding(new_id, repo_url)
```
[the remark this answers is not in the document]
melisande1: this is just for the debug mode, where the if part is made of multiple calls to add_discovery, so I made it similar, this way one can precisely locate where a bug might arise. I don't think there's a significant difference in speed between add_embedding vs add_embeddings, especially since the cursor is not closed

credentialdigger/client.py
```diff
@@ -1039,20 +1202,25 @@ def update_similar_snippets(self,
...
+        # If specified target snippet is not found in the embeddings table,
+        # no update is performed
+        if not target_embedding:
+            return 0
+        # We have a target embedding for the target snippet
+        else:
+            for d in discoveries:
+                if (
+                    d['state'] != state
+                    and self.get_embedding(discovery_id=d['id'])
```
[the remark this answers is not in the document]
melisande1: The issue with that is that we only want to retrieve embeddings that are not from the current state (and the state in not present in the discovery table). Moreover, I didn't make a get_embeddings method, because just getting a list of embeddings without any id would not be very useful, so it would need to return a list of tuples with ids and embeddings. This matching between embeddings and ids is more easily achieved using a get_embedding with the id as argument. And when I see the similarity update being performed rather instantly, even on large repos, I don't think there's a problem in making individual calls to get_embedding, but I'm happy to try otherwise

credentialdigger/client_sqlite.py
```diff
@@ -181,6 +192,60 @@ def add_discoveries(self, discoveries, repo_url):
...
+        repo_url: str
+            The discovery's repository url
+        """
+        snippet = self.get_discovery(discovery_id)['snippet']
+        if not embedding:
+            model = build_embedding_model()
+            embedding = compute_snippet_embedding(snippet, model)
+        embedding_str = ""
+        for emb in embedding:
+            embedding_str += str(emb) + ","
```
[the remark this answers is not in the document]
melisande1: yes I should have used join

credentialdigger/client_postgres.py
melisande1: yes I'm not sure how to specify the list type in the docstring

melisande1:
@marcorosa I made the requested changes. Now I'm ready to make changes to add_embeddings and update_similar_snippets to remove parts of the code from the child UI classes, but I see no simple solution, except to check for get_discoveries's output length in parent class

melisande1:
@marcorosa I moved the two methods from UI child classes to the parent Client class, by checking for output length of get_discoveries. I'm done with the changes.

melisande1:
@marcorosa Hi, I made complementary changes, so as to comply with all of your comments/requirements. Namely I changed the way of retrieving embeddings in update_similar_snippets from get_embedding to get_embeddings (I added a get_embeddings method), and I created a compute_repo_embeddings method so as to refactor add_embeddings (to move additional code to the parent class)

credentialdigger/client.py
```diff
@@ -112,12 +113,75 @@ def add_discovery(self, query, file_name, commit_id, line_number, snippet,
...
         return
 
+    def add_embedding(self,
+                      query,
+                      discovery_id,
+                      snippet,
+                      repo_url,
+                      embedding=None):
+        """ Add an embedding to the embeddings table.
+        Parameters
```
marcorosa, line 130: missing blank line above

credentialdigger/client.py
```diff
@@ -1037,22 +1251,24 @@ def update_similar_snippets(self,
...
             The number of similar snippets found and updated
         """
-
-        discoveries = self.get_discoveries(repo_url, file_name)
-        model = build_embedding_model()
-        """ Compute target snippet embedding """
-        target_snippet_embedding = compute_snippet_embedding(target_snippet,
-                                                             model)
+        disc = self.get_discoveries(repo_url, file_name)
+        discoveries = disc[1] if len(disc) == 2 else disc
```
marcorosa, line 1255: please put a comment because this instruction is not clear

marcorosa: why should this repo have only 2 discoveries?

marcorosa (commented):
Hi @melisande1, thank you, great work. I will review it ASAP. 

I know sqlite has some shortcomings, the purpose behind the request to test serialization is to keep the structure of the sqlite and postgres db consistent as much as possible. So, if you can test it (without pushing your new code before I finish the review) I will very appreciate it, otherwise we can take this into consideration later. But it's an option we need to verify and address before merging into the main branch, since we are doing a major breaking change.
But again, it's just for style and consistency, I know the string hack works fine.

melisande1:
Hi @marcorosa, thank you. Okay, fine I will test serialization (and will not push changes until you're done reviewing the code). I'm assuming you want to apply serialization to both sqlite and postgres, so as to keep them similar. (Initially I had in mind you wanted to make changes only to sqlite, and this would make the dbs even less similar as they are now, but I think I misunderstood)

credentialdigger/client.py
melisande1: That's the part where I'm handling the different output types of get_discoveries in the UI vs regular clients. In the UI classes, the discoveries are given by get_discoveries()[1] (while the discoveries count is givent by get_discoveries()[0]), so checking for output length allows to retrieve the discoveries in case the client is a UI client. I don't think that's a proper way of doing it, but otherwise I need to copy paste the code in the UI classes, which you don't seem to agree with.

credentialdigger/client_postgres.py
```diff
@@ -195,6 +198,48 @@ def add_rule(self, regex, category, description=''):
             query='INSERT INTO rules (regex, category, description) \
                     VALUES (%s, %s, %s) RETURNING id')
 
+    def add_embedding(self, discovery_id, embedding=None, repo_url=''):
```
marcorosa, line 201: Can an embedding have an empty repo url? Isn't it an external key in the db?

credentialdigger/client.py
marcorosa: Ah ok now I understand. So, if this code is specific for the UI and doesn't work with the normal Client, we have to move it to ClientUI. And here keep a method working with the calls of the Client (if really needed).

credentialdigger/client_sqlite.py
```diff
@@ -49,6 +51,14 @@ def __init__(self, path):
                 FOREIGN KEY (rule_id) REFERENCES rules ON DELETE SET NULL ON UPDATE CASCADE
             );
 
+            CREATE TABLE IF NOT EXISTS embeddings (
+                id INTEGER REFERENCES discoveries,
+                snippet TEXT DEFAULT '',
+                embedding TEXT DEFAULT '',
```
marcorosa, line 57: Do we need a default=''? If I remember correctly, NULL values are allowed in the db (unless you define NOT NULL, which is not the case here)

ui/backend/client_ui.py
```diff
@@ -101,49 +101,3 @@ def check_repo(self, repo_url, git_token=None, local_repo=False):
...
-
-        Returns
-        -------
-        int
-            The number of similar snippets found and updated
-        """
-
-        discoveries = self.get_discoveries(repo_url,
-                                           file_name,
-                                           state_filter='new')[1]
```
marcorosa, line 136: now I get why there was this 1 (follow-up from your previous comment in the parent Client class).
So, I'm wondering if we should keep this method here. Or maybe, we can delegate part of this call to Client, and keep in ClientUI the part related to finding the correct discoveries. I can help you on this.

marcorosa (changes requested):
I have a few comments, but I can give you assistance with the `update_similar_snippets` method

marcorosa:
@melisande1  wrt serialization: I was thinking of it only for sqlite. I know postgres support lists (e.g., the list of floats we need for the embeddings). The lack of support is from the sqlite part, so I was thinking on it only for sqlite. But if you think we can be consistent by serializing the embeddings for both the db so to stay consistent (and possibly to have easier maintainability of the code by writing 1 single method for both the clients) then it's worthwhile to try.

ui/backend/client_ui.py
melisande1: yes your help on the matter would be appreciated, thank you, apart from the current one, I can only think of rather complicated solutions

credentialdigger/client_sqlite.py
melisande1: no I tried without, it works as well

credentialdigger/client_postgres.py
melisande1: okay, I'll remove the default value

melisande1:
@marcorosa okay, I see, I'll try with both, then

credentialdigger/client_sqlite.py
```diff
@@ -398,6 +493,67 @@ def get_discovery_group(self, repo_url, state=None):
...
+            query = 'SELECT embedding FROM embeddings WHERE snippet=?'
+        else:
+            return None
+        str_embedding = super().get_embedding(query=query,
+                                              discovery_id=discovery_id,
+                                              snippet=snippet)
+        embedding = []
+        if str_embedding:
+            embedding = [float(emb)
+                         for emb in str_embedding.split(",")]
```
marcorosa, line 531: please fix single ' for strings

credentialdigger/client_sqlite.py
```diff
@@ -398,6 +493,67 @@ def get_discovery_group(self, repo_url, state=None):
...
+            A dictionary with discovery ids as keys and matching
+            embeddings as values
+        """
+        query = 'SELECT id, embedding FROM embeddings WHERE repo_url=?;'
+        str_embeddings_dict = super().get_embeddings(query=query,
+                                                     repo_url=repo_url)
+        embeddings_dict = {}
+        for id, str_embedding in str_embeddings_dict.items():
+            embeddings_dict[id] = [float(emb)
+                                   for emb in str_embedding.split(",")]
```
marcorosa, line 554: please fix single ' for strings

credentialdigger/client_postgres.py
marcorosa: @melisande1 you didn't notice this

marcorosa (changes requested):
Hi, the PR now it's correct and I will merge it, but first I have one very last request: please fix a few style inconsistencies (use `'` instead of `"`), a typo, and the comments from @pep8speaks (there are some left). After these, I will merge

melisande1:
Hi, thank you, I just made those changes

marcorosa (approved):

PR #118, opened 2021-06-14
[Fix] Fixed multiple pre-processing bugs

alaabenfatma:
#### List of fixed bugs:
- BUG: the tool fails to extract the passwords correctly when they contain special characters
    - i.e : password = "hello!all."
        - we get : [password, hello, all]
    - correct: [password, hello!all.]
    [!] Fixed (regex)
 
- BUG: The tool ignores assignments between quotes
    - i.e 
            "Env": [
                "NPM_PASSWORD=123"
            ]
        the tool ignores this because it thinks that it is a 1 word password (because it is contained within two commas)
    - correct result: this should have been considered as an assignment and treated accordingly => [NPM_PASSWORD, 123]
    [!] Fixed (regex)
 
- BUG: In some cases, the tool removes the special characters before extracting the password
    - i.e MSSQL_PWD=Db3xpr3ss10n#F => [MSSQL_PWD, Db3xpr3ss10n, F]
    - correct res: [MSSQL_PWD, Db3xpr3ss10n#F]
    [!] Fixed using regex + strings manipulation techniques

- BUG: the passwords contained between commas get converted into CameCase format
	- i.e pwd = "hello_world" => [Pwd, HelloWorld]
	- [!] Fixed by excluding the words extracted from commas from the converting algorithms.

- BUG: we try to extract private keys and classify their values.
    - Instead of extracting the value of the key, we now that private keys have a certain format to them so we classify them as Leaks directly without feeding them to the classifier.

>

### DIFF
--- a/credentialdigger/models/snippet_model/snippet_model.py
+++ b/credentialdigger/models/snippet_model/snippet_model.py
@@ -76,7 +76,7 @@ class SnippetModel(BaseModel):
             # No need to run the model: we assume this is a false positive
             # since either the snippet is empty or there is just one word
             return True
-        
+
         # Classify as a 'Leak' if this is a private key.
         if self._check_private_key(data):
             return False
@@ -108,8 +108,8 @@ class SnippetModel(BaseModel):
         We define as strings any sequence of characters included in `"` or `'`
         and as words any sequence of alphanumeric characters.
 
-        Finally, convert words (not strings) from snake_case (i.e., words 
-        separated by underscores, like in Python convention) to CamelCase 
+        Finally, convert words (not strings) from snake_case (i.e., words
+        separated by underscores, like in Python convention) to CamelCase
         (i.e., Java convention).
 
         Parameters
@@ -147,9 +147,9 @@ class SnippetModel(BaseModel):
         ]
         """
         # Extract all the words in a snippet
-        words = re.findall("(?<=').*?(?=')|(?<=\").*?(?=\")|[\w\d]+", raw_data)
+        words = re.findall(r'(?<=\').*?(?=\')|(?<=").*?(?=")|[\w\d]+', raw_data)
         # Extract only the words that are between " or ', we refer to them strings
-        strings = re.findall(r"(?<=').*?(?=')|(?<=\").*?(?=\")", raw_data)
+        strings = re.findall(r'(?<=\').*?(?=\')|(?<=").*?(?=")', raw_data)
 
         camel_case_words = []
         for w in words:
@@ -217,7 +217,6 @@ class SnippetModel(BaseModel):
             r'^((\s*|\ *)\@\@.*\@\@(\s*|\ *)|(\s*|\ *)\+(\s*|\ *)|(\s*|\ *)\-(\s*|\ *)|(\s*|\ *))',
             "",
             snippet).strip()
-
 
     def _check_private_key(self, snippet):
         """ Check if this snippet is a private key
--- a/credentialdigger/models/snippet_model/snippet_model.py
+++ b/credentialdigger/models/snippet_model/snippet_model.py
@@ -133,17 +133,19 @@ class SnippetModel(BaseModel):
 
         >>> raw_data = '"password": "#####", "!@AAA12")'
         >>> print(self._pre_process(raw_data))
-        ['password', 'AAA12']
+        ['password','"#####', '!@AAA12']
         """
-        # r"((?<=').*?(?=')|(?<=\").*?(?=\")|[a-zA-Z0-9\._@-]+)"
-        # Match words and strings
-        # In strings, match the content starting from the first character
-        # e.g., "****" -> None
-        #       "$AAA123!!" -> AAA123!!
-        pattern = re.compile(
-            r"((?<=')\w\d.*?(?=')|(?<=\")\w\d.*?(?=\")|[\w\d]+)")
-        words = re.findall(pattern, raw_data)
-        return list(map(string_utils.snake_case_to_camel, words))
+        words = re.findall("(?<=').*?(?=')|(?<=\").*?(?=\")|[\w\d]+", raw_data)
+        strings = re.findall(r"(?<=').*?(?=')|(?<=\").*?(?=\")", raw_data)
+
+        camel_case_words = []
+        for w in words:
+            if(w in strings):
+                camel_case_words.append(w)
+            else:
+                camel_case_words.append(string_utils.snake_case_to_camel(w))
+
+        return camel_case_words
 
     def _label_preprocess(self, words_list):
         """ Output the extracted word from the label of the model.
--- a/credentialdigger/models/snippet_model/snippet_model.py
+++ b/credentialdigger/models/snippet_model/snippet_model.py
@@ -1,10 +1,11 @@
+from difflib import SequenceMatcher
 import re
 
 import fasttext
 import string_utils
 
 from ..base_model import BaseModel
-from difflib import SequenceMatcher
+
 
 EXTENSIONS = set(['py', 'rb', 'c', 'cpp', 'cs', 'js', 'php', 'h', 'java', 'pl',
                   'go'])
@@ -43,9 +44,11 @@ class SnippetModel(BaseModel):
     def analyze(self, discovery):
         """ Analyze a snippet and predict whether it is a false positive or not.
         Parameters
+
         ----------
         discovery: dict
             A discovery
+
         Returns
         -------
         bool
@@ -68,25 +71,24 @@ class SnippetModel(BaseModel):
 
         # Convert data format
         data = self._pre_process(raw_data)
-        # Classify as a 'Leak' if this is a private key.
-        if(self._check_private_key(data)):
-            return False
-        if(not any(not c.isalnum() and c not in ' _.,?!/' for c in raw_data)):
-            # We ignore snippets that look like regular phrases with
-            # no assignment
-            return True
 
         if len(data) < 2:
             # No need to run the model: we assume this is a false positive
             # since either the snippet is empty or there is just one word
-            if(not self._assignment_expression(raw_data)):
-                return True
-            else:
-                data = self._pre_process(raw_data)
+            return True
+        
+        # Classify as a 'Leak' if this is a private key.
+        if self._check_private_key(data):
+            return False
+
+        # We ignore snippets that look like regular phrases with
+        # no assignment
+        if not any(not c.isalnum() and c not in ' _.,?!/' for c in raw_data):
+            return True
 
         input_text = data[-1]
 
-        if(len(input_text) <= 3):
+        if len(input_text) <= 3:
             # We ignore any password shorter than or equal to 3
             return True
         else:
@@ -134,12 +136,16 @@ class SnippetModel(BaseModel):
         >>> print(self._pre_process(raw_data))
         ['password','"#####', '!@AAA12']
         """
+        # Extract all the words in a snippet
         words = re.findall("(?<=').*?(?=')|(?<=\").*?(?=\")|[\w\d]+", raw_data)
+        # Extract only the words that are between " or ', we refer to them strings
         strings = re.findall(r"(?<=').*?(?=')|(?<=\").*?(?=\")", raw_data)
 
+        
         camel_case_words = []
         for w in words:
-            if(w in strings):
+            # If a word is a string, we do not make any changes to it
+            if w in strings:
                 camel_case_words.append(w)
             else:
                 camel_case_words.append(string_utils.snake_case_to_camel(w))
@@ -203,19 +209,10 @@ class SnippetModel(BaseModel):
             "",
             snippet).strip()
 
-    def _assignment_expression(self, snippet):
-        """ Check if the snippet is an assignment or not.
-        """
-        assignment_symbols = ['=', ':=', '<-', ':']
-        for symbol in assignment_symbols:
-            findings = re.findall(f'(.*){symbol}(.*)', snippet)
-            if(len(findings) > 0):
-                return True
-        return False
 
     def _check_private_key(self, snippet):
         """ Check if this snippet is a private key
         """
         base_private_key = ['BEGIN', 'PRIVATE', 'KEY']
-        # Return True if similarity ration >= 85%
+        # Return True if similarity ratio >= 85%
         return SequenceMatcher(None, base_private_key, snippet).ratio() >= 0.85
--- a/credentialdigger/models/snippet_model/snippet_model.py
+++ b/credentialdigger/models/snippet_model/snippet_model.py
@@ -101,7 +101,7 @@ class SnippetModel(BaseModel):
         Find words and strings in `raw_data` and extract them.
         We define as strings any sequence of characters included in `"` or `'`
         and as words any sequence of alphanumeric characters.
-
+        
         In particular, if a string does not contain characters, it is not
         matched. This is done in order to exclude patterns such as "###" or
         "***". In case the string has a mixed content, it is matched starting
--- a/credentialdigger/models/snippet_model/snippet_model.py
+++ b/credentialdigger/models/snippet_model/snippet_model.py
@@ -4,6 +4,7 @@ import fasttext
 import string_utils
 
 from ..base_model import BaseModel
+from difflib import SequenceMatcher
 
 EXTENSIONS = set(['py', 'rb', 'c', 'cpp', 'cs', 'js', 'php', 'h', 'java', 'pl',
                   'go'])
--- a/credentialdigger/models/snippet_model/snippet_model.py
+++ b/credentialdigger/models/snippet_model/snippet_model.py
@@ -210,7 +210,6 @@ class SnippetModel(BaseModel):
         for symbol in assignment_symbols:
             findings = re.findall(f'(.*){symbol}(.*)', snippet)
             if(len(findings) > 0):
-                #print(f'This is an assignment : {findings}')
                 return True
         return False
 
--- a/credentialdigger/models/snippet_model/snippet_model.py
+++ b/credentialdigger/models/snippet_model/snippet_model.py
@@ -141,7 +141,7 @@ class SnippetModel(BaseModel):
 
         >>> raw_data = '"password": "#####", "!@AAA12")'
         >>> print(self._pre_process(raw_data))
-        ['password','"#####', '!@AAA12']
+        ['password','"#####"', '!@AAA12']
 
         Manual execution
         --------
--- a/credentialdigger/models/snippet_model/snippet_model.py
+++ b/credentialdigger/models/snippet_model/snippet_model.py
@@ -135,13 +135,22 @@ class SnippetModel(BaseModel):
         >>> raw_data = '"password": "#####", "!@AAA12")'
         >>> print(self._pre_process(raw_data))
         ['password','"#####', '!@AAA12']
+
+        Manual execution
+        --------
+        >>> raw_data = 'pwd = "pwd_123";'
+        >>> print(self._pre_process(raw_data))
+        [
+            words = [pwd, pwd_123]
+            strings = [pwd_123] # Will not be converted into camel_case
+            camel_case_words = [Pwd, pwd_123]
+        ]
         """
         # Extract all the words in a snippet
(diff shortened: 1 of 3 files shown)


credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -4,6 +4,7 @@
 import string_utils
 
 from ..base_model import BaseModel
+from difflib import SequenceMatcher
```
marcorosa, line 7: this is part of the standard library, so it shouldn't be placed here [ref](https://google.github.io/styleguide/pyguide.html#313-imports-formatting)

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -41,19 +42,16 @@ def __init__(self,
 
     def analyze(self, discovery):
         """ Analyze a snippet and predict whether it is a false positive or not.
-
```
marcorosa, line 44: leave a blank space here (style rules)

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -41,19 +42,16 @@ def __init__(self,
...
         Parameters
         ----------
         discovery: dict
             A discovery
-
```
marcorosa, line 49: leave a blank space (style rules)

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -70,23 +68,29 @@ def analyze(self, discovery):
 
         # Convert data format
         data = self._pre_process(raw_data)
+        # Classify as a 'Leak' if this is a private key.
+        if(self._check_private_key(data)):
```
marcorosa, line 72: no () appended to the `if`. This should have been already raised by @pep8speaks, but it didn't.

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -201,3 +202,20 @@ def _remove_initial_junk(self, snippet):
...
+            findings = re.findall(f'(.*){symbol}(.*)', snippet)
+            if(len(findings) > 0):
+                return True
+        return False
+
+    def _check_private_key(self, snippet):
+        """ Check if this snippet is a private key
+        """
+        base_private_key = ['BEGIN', 'PRIVATE', 'KEY']
+        # Return True if similarity ration >= 85%
```
marcorosa, line 220: typo in ratio

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -70,23 +68,29 @@ def analyze(self, discovery):
 
         # Convert data format
         data = self._pre_process(raw_data)
+        # Classify as a 'Leak' if this is a private key.
+        if(self._check_private_key(data)):
+            return False
+        if(not any(not c.isalnum() and c not in ' _.,?!/' for c in raw_data)):
```
marcorosa, line 74: what is this check?

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -70,23 +68,29 @@ def analyze(self, discovery):
...
+            # We ignore snippets that look like regular phrases with
+            # no assignment
+            return True
 
         if len(data) < 2:
             # No need to run the model: we assume this is a false positive
             # since either the snippet is empty or there is just one word
-            return True
+            if(not self._assignment_expression(raw_data)):
```
marcorosa, line 82: same comment on `if`

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -201,3 +202,20 @@ def _remove_initial_junk(self, snippet):
...
             "",
             snippet).strip()
+
+    def _assignment_expression(self, snippet):
+        """ Check if the snippet is an assignment or not.
+        """
+        assignment_symbols = ['=', ':=', '<-', ':']
+        for symbol in assignment_symbols:
+            findings = re.findall(f'(.*){symbol}(.*)', snippet)
+            if(len(findings) > 0):
```
marcorosa, line 212: same comment on `if`

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -201,3 +202,20 @@ def _remove_initial_junk(self, snippet):
             r"^((\s*|\ *)\@\@.*\@\@(\s*|\ *)|(\s*|\ *)\+(\s*|\ *)|(\s*|\ *)\-(\s*|\ *)|(\s*|\ *))",
             "",
             snippet).strip()
+
+    def _assignment_expression(self, snippet):
+        """ Check if the snippet is an assignment or not.
+        """
+        assignment_symbols = ['=', ':=', '<-', ':']
+        for symbol in assignment_symbols:
+            findings = re.findall(f'(.*){symbol}(.*)', snippet)
```
marcorosa, line 211: isn't it sufficient a piece of code like this?

```
for symbol in symbols:
    return True if symbol in snippet
return False
```

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -201,3 +202,20 @@ def _remove_initial_junk(self, snippet):
             r"^((\s*|\ *)\@\@.*\@\@(\s*|\ *)|(\s*|\ *)\+(\s*|\ *)|(\s*|\ *)\-(\s*|\ *)|(\s*|\ *))",
             "",
             snippet).strip()
+
+    def _assignment_expression(self, snippet):
+        """ Check if the snippet is an assignment or not.
+        """
+        assignment_symbols = ['=', ':=', '<-', ':']
```
marcorosa, line 209: is `<-` correct? isn't it on the other way?

`myclass -> myfield = "somevalue"`

marcorosa: moreover, I wonder if we are missing some other way to assign values. E.g., `strcpy` calls in c

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -70,23 +68,29 @@ def analyze(self, discovery):
...
+                data = self._pre_process(raw_data)
 
-        if len(data) == 2:
-            # No need to pre-process data since we have only two words
-            # Predict if the string `word1 + ' ' + word2` is a false positive
-            label = self.model.predict(
-                data[0] + ' ' + data[1])[0]  # 0=label, 1=probability
+        input_text = data[-1]
+
+        if(len(input_text) <= 3):
```
marcorosa, line 89: same comment on `if`

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -133,17 +132,19 @@ def _pre_process(self, raw_data):
...
-        # Match words and strings
-        # In strings, match the content starting from the first character
-        # e.g., "****" -> None
-        #       "$AAA123!!" -> AAA123!!
-        pattern = re.compile(
-            r"((?<=')\w\d.*?(?=')|(?<=\")\w\d.*?(?=\")|[\w\d]+)")
-        words = re.findall(pattern, raw_data)
-        return list(map(string_utils.snake_case_to_camel, words))
+        words = re.findall("(?<=').*?(?=')|(?<=\").*?(?=\")|[\w\d]+", raw_data)
+        strings = re.findall(r"(?<=').*?(?=')|(?<=\").*?(?=\")", raw_data)
```
marcorosa, line 138: use `'` to define strings, and `"` for the docstrings (I know these strings were already defined with `"`, but please change it because it's not consistent with the rest of the codebase)

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -133,17 +132,19 @@ def _pre_process(self, raw_data):
...
+
+        camel_case_words = []
+        for w in words:
+            if(w in strings):
+                camel_case_words.append(w)
+            else:
+                camel_case_words.append(string_utils.snake_case_to_camel(w))
+
+        return camel_case_words
```
marcorosa, line 147: Please write a comment otherwise this piece of code is not comprehensible.

credentialdigger/models/snippet_model/snippet_model.py
marcorosa: what's the difference between strings and words? What should they contain?

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -201,3 +202,20 @@ def _remove_initial_junk(self, snippet):
             r"^((\s*|\ *)\@\@.*\@\@(\s*|\ *)|(\s*|\ *)\+(\s*|\ *)|(\s*|\ *)\-(\s*|\ *)|(\s*|\ *))",
             "",
```
marcorosa, line 203: please replace " with ' for these strings (we were not consistent before this PR but I didn't notice it)

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -133,17 +132,19 @@ def _pre_process(self, raw_data):
...
-        pattern = re.compile(
-            r"((?<=')\w\d.*?(?=')|(?<=\")\w\d.*?(?=\")|[\w\d]+)")
-        words = re.findall(pattern, raw_data)
-        return list(map(string_utils.snake_case_to_camel, words))
+        words = re.findall("(?<=').*?(?=')|(?<=\").*?(?=\")|[\w\d]+", raw_data)
+        strings = re.findall(r"(?<=').*?(?=')|(?<=\").*?(?=\")", raw_data)
+
+        camel_case_words = []
+        for w in words:
+            if(w in strings):
```
marcorosa, line 142: same comment on the if

marcorosa (changes requested):
There are several improvements needed before accepting this PR.
Moreover, please follow the indications of @pep8speaks to improve basic style errors (it should be the first message of this PR since it's a bot)

credentialdigger/models/snippet_model/snippet_model.py
alaabenfatma: As the comments below indicates, it is used to ignore snippets that look like regular phrases.
Example : 
```python
# Insert you password below
```
Will be ignored and not passed to the classifier. In other words, it filters out mostly comments and strings that get detected by the regex rules and are not containing password/token assignments.

credentialdigger/models/snippet_model/snippet_model.py
alaabenfatma: > isn't it sufficient a piece of code like this?
> 
> ```
> for symbol in symbols:
>     return True if symbol in snippet
> return False
> ```

Yes, at first I wanted to extract the findings using ReGex but then I did not follow that path so I ended up only checking for the existence of the symbols. will replace the code.

credentialdigger/models/snippet_model/snippet_model.py
alaabenfatma: > what's the difference between strings and words? What should they contain?

snippet = `pwd = "123"` 
words = [pwd, 123]
strings = [123]

We will only convert that words that are not contained between " or ' into camel_case.

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -76,17 +77,24 @@ def analyze(self, discovery):
...
+        index_of_value = self._label_preprocess(data)
+        
+        # Classify as a 'Leak' if this is a private key.
+        if self._check_private_key(data):
+            return False
+
+        # We ignore snippets that look like regular phrases with
+        # no assignment
+        if not any(not c.isalnum() and c not in ' _.,?!/' for c in raw_data):
+            return True
```
marcorosa, line 90: I don't understand this line. What are you checking? That there is at least 1 symbol, and that this symbol is not in `' _.,?!/'` ?

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -76,17 +77,24 @@ def analyze(self, discovery):
             # since either the snippet is empty or there is just one word
             return True
 
-        if len(data) == 2:
-            # No need to pre-process data since we have only two words
-            # Predict if the string `word1 + ' ' + word2` is a false positive
-            label = self.model.predict(
-                data[0] + ' ' + data[1])[0]  # 0=label, 1=probability
+        # Extract the committed secret
+        index_of_value = self._label_preprocess(data)
```
marcorosa, line 81: Place a comment here to keep track of the format of this variable, otherwise we'll forget it (more on this in the command of L92)

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -76,17 +77,24 @@ def analyze(self, discovery):
...
+        # Classify as a 'Leak' if this is a private key.
+        if self._check_private_key(data):
+            return False
+
+        # We ignore snippets that look like regular phrases with
+        # no assignment
+        if not any(not c.isalnum() and c not in ' _.,?!/' for c in raw_data):
+            return True
+
+        input_text = data[index_of_value[1]]
```
marcorosa, line 92: place a comment here explaining why you do this apparently weird transformation: why position 1 of `index_of_value`? Why is data needed?

marcorosa: `input_text` is a misleading name: this is not an input, this is "a bet", i.e., the value that the extractor identify as the secret/token/password value

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -76,17 +77,24 @@ def analyze(self, discovery):
...
+        if len(input_text) <= 3:
+            # We ignore any password shorter than or equal to 3
+            return True
         else:
-            finish_labels = self._label_preprocess(data)
-            # Predict if the string `word1 + ' ' + word2` is a false positive
```
marcorosa, line 86: leave this comment to explain why we need to launch a ML prediction here

marcorosa (changes requested):

credentialdigger/models/snippet_model/snippet_model.py
alaabenfatma: I am checking if the snippet at hand is a normal English expression and cannot (in almost all cases) be containing a leak.
Example:
```python
# put your password here
pwd=""
```
Using that check, we will classify the comment as FP without feeding it to the classifier.
NB: I know that comments may contain leaks, but I am not excluding comments, am excluding sentences that do not usually fall under code snippets that may contain leaks.

A real example: 
```python
raw_data = 'This is a code snippet, it contains no password. words like api_token, pwd, pass will make it detectable via the regex scanner, but this check tell us not to bother with. We know for sure that it is an FP! will this make things faster when scanning? Yes! . __copyrights__!.\
    Source file is /src/fake.txt'
if not any(not c.isalnum() and c not in ' _.,?!/' for c in raw_data):
    print(True)
else:
    print(False)
```
This code will print(True) => FP without us feeding this to the classifier. (
I have used all the mentioned symbols within the `raw_data` string.

credentialdigger/models/snippet_model/snippet_model.py
alaabenfatma: > place a comment here explaining why you do this apparently weird transformation: why position 1 of index_of_value? Why is data needed?

- data is what we are left with after pre-processing the snippet
   - if we have `raw_data='pass_word="123"`, then data will be `[PassWord, 123]`
- we use only position 1 because the new model (and then old one too, that's been trained using the new datasets (to be eventually released)) only takes 1 value as input, contrary to the old strategy of taking a couple (key,value). So we only retrieve the value (index 1) and ignore the variable name (index 0).

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -198,6 +214,13 @@ def _remove_initial_junk(self, snippet):
         """ Remove junk from the beginning of a snippet.
         """
         return re.sub(
-            r"^((\s*|\ *)\@\@.*\@\@(\s*|\ *)|(\s*|\ *)\+(\s*|\ *)|(\s*|\ *)\-(\s*|\ *)|(\s*|\ *))",
+            r'^((\s*|\ *)\@\@.*\@\@(\s*|\ *)|(\s*|\ *)\+(\s*|\ *)|(\s*|\ *)\-(\s*|\ *)|(\s*|\ *))',
             "",
             snippet).strip()
+
+    def _check_private_key(self, snippet):
+        """ Check if this snippet is a private key
```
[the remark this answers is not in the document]
alaabenfatma: > Moreover, you're not checking if the snippet is a private key itself, but only if it's the header of a private key

Yes, we are only checking if this is a header of a private key. Limited by the performance of the Extractor, it never did successfully extract correctly any private key. That said, we now are only testing if we are dealing with a private key's header or not.

On the other hand, I am not in favor of using the Extractor/Classifier for this one task. I believe that it would be better to use a rule such as [`-----BEGIN (\s|.*)PRIVATE KEY-{3,}\n([\s\S]*?)\n-{3,}END (\s|.*)PRIVATE KEY-----`](https://regex101.com/r/zyFPxJ/1) to find out if we are dealing with a private key or not. I made this regex to detect:
- the type of the private key (i.e RSA)
- the contents

We have to find a way to tell that we found a private key, then instead of passing it to the snippet model, we classify it directly as an FP.

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -133,17 +141,32 @@ def _pre_process(self, raw_data):
 
         >>> raw_data = '"password": "#####", "!@AAA12")'
         >>> print(self._pre_process(raw_data))
-        ['password', 'AAA12']
+        ['password','"#####', '!@AAA12']
+
+        Manual execution
+        --------
```
marcorosa, line 147: Too short

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -133,17 +141,32 @@ def _pre_process(self, raw_data):
...
+        >>> raw_data = 'pwd = "pwd_123";'
+        >>> print(self._pre_process(raw_data))
+        [
+            words = [pwd, pwd_123]
+            strings = [pwd_123] # Will not be converted into camel_case
+            camel_case_words = [Pwd, pwd_123]
+        ]
```
marcorosa, line 154: The `_pre_process` method doesn't return this list. If you want to give a good description of these steps, you can do it in the notes (more on this in the final comment of the review)

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -198,6 +221,19 @@ def _remove_initial_junk(self, snippet):
...
+
+    def _check_private_key(self, snippet):
+        """Check if this snippet is a private key
+
+        Args:
+            snippet (str): A code snippet containing a secret
+
+        Returns:
+            boolean: True if this is a header of a private key, False otherwise.
+        """
```
marcorosa, line 236: The style of this doctoring is not consistent with the rest on the codebase (more on this in the final comment)

marcorosa (changes requested):
Please follow the style rules (pep8 and numpy docstrings) of the rest of the codebase. Like it or not, we must be consistent with ourselves.

[Numpy docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html)

Moreover, there is also a bot (@pep8speaks ) that helps you finding style errors. It's commenting all the PRs. Please follow its suggestions.

Finally, the code is not passing the tests. Please verify why: we cannot merge code that is breaking something else.

credentialdigger/models/snippet_model/snippet_model.py
```diff
@@ -198,6 +213,46 @@ def _remove_initial_junk(self, snippet):
...
         """
         return re.sub(
-            r"^((\s*|\ *)\@\@.*\@\@(\s*|\ *)|(\s*|\ *)\+(\s*|\ *)|(\s*|\ *)\-(\s*|\ *)|(\s*|\ *))",
+            r'^((\s*|\ *)\@\@.*\@\@(\s*|\ *)|(\s*|\ *)\+(\s*|\ *)|(\s*|\ *)\-(\s*|\ *)|(\s*|\ *))',
             "",
             snippet).strip()
+
+    def _not_an_assignment(self, snippet):
+        """We ignore snippets that look like regular phrases with no assignment hints.
+        Parameters:
```
alaabenfatma, line 222: string Docs, to be edited.

marcorosa (approved):

PR #122, opened 2021-06-21
Fix/show on GitHub + requirments.txt adaptation

alaabenfatma:
### Fix Show on Github button + update requirments.txt
- It is now possible to "Show on Github" for repos that with urls that end with `.git` 
- Updated the requirments.txt file so it install the adapted version of tensorflow and its sub-libraries according to the installed version of Python.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,7 +13,9 @@ pyyaml
 requests
 scikit-learn
 srsly
-tensorflow==2.4.1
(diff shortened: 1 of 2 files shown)


marcorosa (changes requested):
Please merge `main` into your branch (there are some differences in `requirements.txt`) and resolve the conflicts, then I'll approve

marcorosa (approved):

PR #123, opened 2021-06-21
adjust dependencies

marcorosa:
Different platforms (linux vs darwin) and python versions (notably, v3.6 and 3.7 vs 3.8 and 3.9) need different versions for some packages. Try to automatically manage this

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 fasttext
 GitPython
-hyperscan==0.1.5
+hyperscan==0.2.0; sys.platform == "dawin"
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,6 +1,7 @@
 fasttext
 GitPython
-hyperscan==0.1.5
+hyperscan==0.2.0; sys.platform == "dawin"
```
alaabenfatma, line 3: @marcorosa typo: `darwing` instead of "dawin"

marcorosa: yes, thanks!

PR #126, opened 2021-06-22
Revert "scan snapshots feature"

marcorosa:
Reverts SAP/credential-digger#121


marcorosa:
My fault! 
We need to integrate with other features in `develop` first

Issue #114, opened 2021-06-09
export leaks


marcorosa:
Add a button to export discoveries of a repo (only the leaks) in a csv/excel file

marcorosa:
Idea: 
drop down button to chose 
- only "new" discoveries
- all discoveries (also FPs, not relevant and addressing)

Pay attention to:
- excel trying to evaluate the snippet
- different locale have different cell separators

marcorosa:
Fixed in #194

PR #138, opened 2021-07-07
[Feat] Export discoveries

alaabenfatma:
### Export leaks
This PR targets this issue: https://github.com/SAP/credential-digger/issues/114

### Features

- Ability to choose what discoveries to export
   - Leaks
   - False positives
   - 'Addressing' discoveries
   - 'irrelevant' discoveries
   - Fixed discoveries
- Export leaks as a .csv file


ui/server.py
```diff
@@ -417,6 +424,63 @@ def update_discovery_group():
         return 'OK', 200
 
 
+def assign_categories(discoveries):
```
marcorosa, line 427: please turn this method to "private", i.e., call it `_assign_categories`. In python there is no visibility like public/protected/private in Java and other programming languages, but a `_` before the name of the method is used by developer to indicate it's "private"

ui/server.py
```diff
@@ -417,6 +424,63 @@ def update_discovery_group():
...
+        List of discoveries with categories assigned to them
+    """
+    rulesdict, cat = _get_rules()
+    for discovery in discoveries:
+        if discovery['rule_id']:
+            discovery['category'] = rulesdict[discovery['rule_id']]['category']
+        else:
+            discovery['category'] = '(rule deleted)'
+
+    return discoveries
```
marcorosa, line 447: I think this is not really needed, since lists and dictionaries are data structures managed by reference. So, if you add a new keypair in the function, it should also update the main data structure in the "main" method (e.g., the sort function)

ui/server.py
```diff
@@ -417,6 +424,63 @@ def update_discovery_group():
...
+        states = ['new', 'false_positive',
+                  'addressing', 'not_relevant', 'fixed']
+    else:
+        states = request.form.getlist('check')
+
+    # filter out based on states
+    filtered_discoveries = []
+    for d in discoveries:
+        if d['state'] in states:
+            filtered_discoveries.append(d)
```
marcorosa, line 470: `filtered_discoveries = list(filter(lambda d: d.get('state') in states, discoveries))`

ui/server.py
```diff
@@ -417,6 +424,63 @@ def update_discovery_group():
...
+
+    si = io.StringIO()
+    keys = discoveries[0].keys()
```
marcorosa, line 473: if you don't have discoveries you raise an IndexError here

ui/server.py
```diff
@@ -417,6 +424,63 @@ def update_discovery_group():
...
+    else:
+        states = request.form.getlist('check')
+
+    # filter out based on states
+    filtered_discoveries = []
+    for d in discoveries:
+        if d['state'] in states:
+            filtered_discoveries.append(d)
+
+    si = io.StringIO()
```
marcorosa, line 472: what's this? `si` is not a very good name (and it's prone to be confused with `is`)

ui/server.py
```diff
@@ -1,5 +1,7 @@
+import csv
 import os
 import sys
+import io
```
marcorosa, line 4: not in the correct order

ui/server.py
```diff
@@ -417,6 +424,63 @@ def update_discovery_group():
...
+
+    # filter out based on states
+    filtered_discoveries = []
+    for d in discoveries:
+        if d['state'] in states:
+            filtered_discoveries.append(d)
+
+    si = io.StringIO()
+    keys = discoveries[0].keys()
+    cw = csv.DictWriter(si, keys)
```
marcorosa, line 474: also herre, `cw` is not a good name

ui/server.py
```diff
@@ -417,6 +424,63 @@ def update_discovery_group():
...
+    cw.writeheader()
+    cw.writerows(filtered_discoveries)
+    response_csv = make_response(si.getvalue())
+    response_csv.headers["Content-Disposition"] = "attachment; \
```
marcorosa, line 478: remember to use ' for the strings and " for the docstrings

ui/templates/discoveries/discoveries.html
```diff
@@ -66,15 +74,16 @@ <h3 class="scanDescription"><span id="discoveriesCounter">{{discoveries_count}}<
   <tbody></tbody>
 </table>
 
-
+{% include 'shared/_exportCSVModal.html' %}
```
marcorosa, line 77: should be in the homepage and not in the discoveries view

marcorosa (changes requested):

ui/templates/shared/_exportCSVModal.html
```diff
@@ -0,0 +1,35 @@
+<!-- Export discoveries popup -->
+<div id="exportDiscovereisModal" class="modal">
```
marcorosa, line 2: typo

marcorosa (approved):

Issue #142, opened 2021-07-09
Cannot sort discoveries by "category" nor "snippet"


alaabenfatma:
It enters an infinite loop.

marcorosa:
Can you give more context, please?
Is it an issue that happens with both the clients? How can we reproduce it (e.g., a script to test this bug)? Did you test it with a repo giving only a few discoveries (so that it's very easy to see that it enters an infinite loop)?

alaabenfatma:
> Can you give more context, please?
> Is it an issue that happens with both the clients? How can we reproduce it (e.g., a script to test this bug)? Did you test it with a repo giving only a few discoveries (so that it's very easy to see that it enters an infinite loop)?

Hi @marcorosa, I updated the description

marcorosa:
Ok, but can you be more specific, please? 😄 
At first I though it was related to the data structure we are using (could have been not sortable). Yet, I sorted manually some discoveries and it works fine
```
conn = sqlite3.connect('data.db')
cursor = conn.cursor()
repo_url = '...'
cursor.execute('select * from discoveries where repo_url=?', (repo_url,))
discoveries = cursor.fetchall()
sorted(discoveries, key=lambda d:d[4])  # 4 is the snippet position
```

So, where is this issue coming from? what methods are you using? are you trying to do this sort with the python library or directly in-browser with javascript? are you trying to sort all the discoveries or a repo or just the paginated ones?

alaabenfatma:
Thank you for your comment @marcorosa 
No, it is not related to the data structure nor the database.

> So, where is this issue coming from?

The issue is coming from the `/get_discoveries` route/function. Precisely this line: https://github.com/SAP/credential-digger/blob/c4b000d6994fffc7e9de2ce48d3fbb47516d4be2/ui/server.py#L393

> are you trying to do this sort with the python library or directly in-browser with javascript?

The bug occurs when we try to sort the discoveries on the browser (javascript).

The bug can be reproduced manually without writing a script, by sorting the discoveries via the UI using these built-in buttons:
![image](https://user-images.githubusercontent.com/9027148/125056621-b1f9d580-e0a8-11eb-8d23-d8143fb21822.png)

> are you trying to sort all the discoveries or a repo or just the paginated ones?

I tried both 'all discoveries' and only the paginated ones, but none of them has worked.





### Fix
#### Problem
Sorting discoveries (manually, from the UI, using the built-in sorting buttons), results in this error:
``` 
Traceback (most recent call last):
[...]

PR #143, opened 2021-07-09
[Fix] Quick fix for discoveries sorting

alaabenfatma:
This PR fixes the issue regarding sorting discoveries on the browser discussed here: https://github.com/SAP/credential-digger/issues/142#issuecomment-877055172


alaabenfatma:
The bug occurs also when sorting based on STATE. 
--> will close PR until fully fixed.

Issue #131, opened 2021-06-23
Change the sum of number of discoveries 


SlimTrabelsi:
The current number of discoveries is taking into account FPs .. It should only take TP

marcorosa:
Reopening the issue because the counter is wrong in the "discoveries in file" view (probably not implemented in this view)

PR #155, opened 2021-07-27
[UI] Show true positive sum alongside the total number of discoveries

alaabenfatma:
This PR targets this issue: https://github.com/SAP/credential-digger/issues/131

Modifications:

- We can now see the sum of TP alongside the total discoveries for every repo line.
- We modified the way we calculate the total number of discoveries for all scanned repos.

### DIFF
--- a/ui/res/js/discoveries.js
+++ b/ui/res/js/discoveries.js
@@ -146,7 +146,6 @@ function initDiscoveriesDataTable() {
         },
         dataSrc: function (json) {
           postfix = " Leaks";
-          console.log(json.stateFilter);
           switch (json.stateFilter) {
             case "false_positive":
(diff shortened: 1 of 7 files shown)


ui/res/js/discoveries.js
```diff
@@ -143,6 +145,38 @@ function initDiscoveriesDataTable() {
...
+            case "false_positive":
+              postfix = " False positives";
+              break;
+            case "addressing":
+              postfix = " addressing";
+              break;
+            case "not_relevant":
+              postfix = " not relevant";
+            default:
+              break;
```
marcorosa: either the tags are all capitalized or not

marcorosa (approved):

PR #161, opened 2021-07-30
[File view: Fix] Fix infinite loop when viewing discoveries by file.

alaabenfatma:
The page enters an infinite loop when viewing discoveries from a single file view.

We now check the existence of the HTML elements, if they are **null**, we abort the operation.


marcorosa (approved):

marcorosa:
This PR should solve #163 
To be merged, run, and published next week.

Issue #145, opened 2021-07-20
add get_discoveries to CLI


marcorosa:
In the CLI we have the possibility to add rules and to do scans (`scan`, `scan_user`, `scan_wiki`, `scan_path`, `scan_snapshot`) but we are missing a command to get the results (even if it's always possible to call directly the UI, but this is not immediate for many users)

marcorosa:
Fixed in #194

PR #164, opened 2021-08-02
[Feat] Get discoveries of scanned repo from the CLI

alaabenfatma:
This PR targets this issue: https://github.com/SAP/credential-digger/issues/145

### Intro
It is now possible to retrieve discoveries of a certain repo from the CLI

#### Print discoveries on the terminal
In order to print the discoveries on the terminal in a tabular form, please run this command:
`python -m credentialdigger get_discoveries repo_url --sqlite data.db`

ℹ️If the repo contains more than 30 discoveries, a message will be shown and the user will have the possiblity to export the discoveries instead of printing them on the terminal.

#### Export discoveries as CSV
In order to export the discoveries as a CSV file, without printing them on the terminal, run this command:
`python -m credentialdigger get_discoveries repo_url --sqlite data.db --save csv_file_path.csv`


credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,158 @@
...
+        '--save', default=None, type=str,
+        help='Path of the .csv file to which we export the discoveries.')
+
+
+def print_discoveries(discoveries, repo_url):
+    with console.status(f'[bold]Processing {len(discoveries)} discoveries...'):
+        discoveries_list = pd.DataFrame(discoveries)
+        del discoveries_list['repo_url']
+        del discoveries_list['timestamp']
+        del discoveries_list['rule_id']
```
marcorosa, line 68: leave a comment above saying that repo_url will not be shown since it's passed as argument.
Also timestamp and and rule_id will not be shown because leaks can be identified also without those information

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Convert `int` columns to `str` to be eventually rendered.
+        discoveries_list['id'] = discoveries_list['id'].astype(str)
+        int_to_str = discoveries_list['line_number'].astype(str)
+        discoveries_list['line_number'] = int_to_str
```
marcorosa, line 73: why do you split this operation in 2? can't you do it in one line like the one above?

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,158 @@
...
+
+        for c in columns:
+            table.add_column(c)
+        for row in discoveries_list:
+            table.add_row(*row)
+
+        console.print(table)
+
+
+def export_csv(repo_url, client, save=False):
```
marcorosa, line 92: missing documentation

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    Retrieve discoveries of a git repository.
+
+    Parameters
+    ----------
+    client: `credentialdigger.Client`
+        Instance of the client from which we retrieve results
+    args: `argparse.Namespace`
+        Arguments from command line parser.
+
```
marcorosa, line 129: blank line

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,158 @@
...
+    if len(discoveries) == 0:
+        # if repo has no discoveries, exit
+        console.print(f'[bold] {args.repo_url} has 0 discoveries.')
+    elif len(discoveries) > MAX_NUMBER_DISCOVERIES:
+        response = ''
+        while response.upper() not in ['Y', 'N']:
+            response = console.input(
+                f'[bold]This repository has more than {MAX_NUMBER_DISCOVERIES} \
+discoveries, export them as .csv instead? (Y/N)')
+        if(response.upper() == 'N'):
```
marcorosa, line 151: don't use parentheses here

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,158 @@
...
+    if args.save is not None:
+        export_csv(args.repo_url, client, save=args.save)
+        return True
+
+    if len(discoveries) == 0:
+        # if repo has no discoveries, exit
+        console.print(f'[bold] {args.repo_url} has 0 discoveries.')
+    elif len(discoveries) > MAX_NUMBER_DISCOVERIES:
+        response = ''
+        while response.upper() not in ['Y', 'N']:
```
marcorosa, line 147: add also "yes" and "no"

credentialdigger/client.py
```diff
@@ -1378,3 +1381,64 @@ def update_similar_snippets(self,
                 self.update_discovery(d['id'], state)
                 n_updated_snippets += 1
         return n_updated_snippets
+
+    def assign_categories(self, discoveries):
```
marcorosa, line 1385: this is not a relevant for the Client. Why should we add it? It's only related to the UI rendering, that's why it's in `server.py`

credentialdigger/client.py
```diff
@@ -1378,3 +1381,64 @@ def update_similar_snippets(self,
...
+        ----------
+        repo_url: str
+            The repo containing the discoveries
+        sates: str | list
+            - str: if it equals 'all', then return all discoveries.
+                   return chosen state otherwise (i.e 'false_positive')
+            - list: return all the discoveries that have states contained
+                    in this list (i.e ['new', 'false_positive'])
+
+        Returns
```
marcorosa, line 1415: missing separator -----

credentialdigger/client.py
```diff
@@ -1378,3 +1381,64 @@ def update_similar_snippets(self,
...
+        str
+            Discoveries in CSV format.
```
marcorosa, line 1417: CSV format? you wrote str above. Is str the correct type?

credentialdigger/client.py
```diff
@@ -1378,3 +1381,64 @@ def update_similar_snippets(self,
...
+        """
+        rulesdict = self.get_rules()
+        for discovery in discoveries:
+            if discovery['rule_id']:
+                category = rulesdict[discovery['rule_id'] - 1]['category']
+                discovery['category'] = category
+            else:
+                discovery['category'] = '(rule deleted)'
+
+    def export_discoveries_csv(self, repo_url, states='all'):
```
marcorosa, line 1402: why should we add this method here while it is introduced only for a CLI?
Can't you implement it in the CLI?

marcorosa (changes requested):
First of all, `get_discoveries` has an optional argument (`file_name`) that has not been implemented.

Then, I would find useful having a `--group` optional argument that calls `get_discoveries_group` instead of `get_discoveries`

Finally, I have several questions on the code and on the functions introduced in the Client class

credentialdigger/client.py
alaabenfatma: It has been added here because it is being called by the export_discoveries_csv

credentialdigger/client.py
alaabenfatma: Yes, str is the correct return type. 
CSV (comma-separated values) is a string itself, it is a string that has a certain format. So in the end, I believe that it is safe to call CSV a string.

credentialdigger/client.py
alaabenfatma: This function has been added here because I am willing to use it later on for the 'export_discoveries' feature requested by this issue here: https://github.com/SAP/credential-digger/issues/114

I wanted to to use one export function for these two different features, and I found the Client to be the best place to contain such a function.

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+The 'get_discoveries' module can be used to retrieve the discoveries on the fly
+from the terminal. It supports both the Sqlite and Postgres clients.
+
+NOTE: Postgres is used by default. Please make sure that the environment
+variables are exported.
+
+usage: credentialdigger get_discoveries [-h] [--dotenv DOTENV]
+                                        [--sqlite SQLITE]
+                                        [--save SAVE]
+                                        repo_url
```
marcorosa, line 11: from below I see there are 2 more options that you haven't listed here: filename and state

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+
+optional arguments:
+    -h, --help          Show this help message and exit
+    --dotenv DOTENV     The path to the .env file which will be used in all
+                        commands. If not specified, the one in the current
+                        directory will be used (if present)
+    --sqlite SQLITE     If specified, scan the repo using the sqlite client
+                        passing as argument the path of the db. Otherwise, use
+                        postgres (must be up and running)
+    --filename FILENAME The filename to filter discoveries on
```
marcorosa, line 25: `Show only the discoveries contained in this file` would be a better description

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+"""
+
+import csv
+import io
+import logging
+import os
+import sys
+
+import pandas as pd
+
```
marcorosa, line 40: why is there this blank line?

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+from rich.console import Console
```
marcorosa, line 41: if you add a dependency, remember to add it also in `requirements.txt` and to check the file with the licenses (for SAP compliancy)

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+        console.print(table)
+
+
+def discoveries_to_csv(discoveries):
+    """ Generate CSV from list of discoveries
+
+    Parameters
+    ----------
+        discoveries: list
+            List of discoveries from which to generate the CSV
```
marcorosa, line 124: why is there an indentation?

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+
+    Returns
+    -------
+        str
```
marcorosa, line 128: again, over indented

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+            A string containing CSV obtaining from the original list of
```
marcorosa, line 129: typo? -> obtained

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+
+    Parameters
+    ----------
+        discoveries: list
+            List of discoveries from which to generate the CSV
+        client: `credentialdigger.Client`
+            Instance of the client from which we retrieve rules
+        save: bool
+            If True, we do not ask the user to enter a file path for the CSV
+            to be exported
```
marcorosa, line 157: as before, you're fighting with over indentation

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+    """
+    # Check if --save is specified
+    if save is False:
+        path = ''
+        # Read the export path from the console's input
+        while path == '':
+            path = console.input('Path to export CSV:')
```
marcorosa, line 164: if `save` is false, then why shall we specify a path? If save is false, then no path is needed

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+    try:
+        with open(path, newline='', mode='w') as csv_file:
+            with console.status('[bold]Exporting the discoveries..'):
+                # Add the category to each discovery
+                assign_categories(client, discoveries)
+                data = discoveries_to_csv(discoveries)
+                csv_file.writelines(data)
+                console.print(
+                    '[bold][!] The discoveries have been exported \
+successfully.')
```
marcorosa, line 178: does `console.print` works like a normal `print`? Therefore, is it possible to format these 2 lines like
```
console.print(
        '[bold][!] The discoveries have been exported '
        'successfully.'
)
```
?

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+    except Exception as e:
```
marcorosa, line 179: What exceptions may be raised? Are there really so many ones that we need to catch directly `Exception`?

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+        console.print(f'[red]{e}[/]')
+        try:
+            os.remove(path)
+        except OSError as osE:
+            console.print(f'[red]{osE}[/]')
```
marcorosa, line 184: a try-except inside an except??

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+
+    Parameters
+    ----------
+    discoveries: list
+        List of discoveries to be filtered
+    states: str | list
+            - str: if it equals 'all', then return all discoveries.
+                   return chosen state otherwise (i.e 'false_positive')
+            - list: return all the discoveries that have states contained
+                    in this list (i.e ['new', 'false_positive'])
```
marcorosa, line 217: it's over indented, and it's better to write a description than leaving the items like this.
Moreover, I suggest to export either all or only 1 state. So the only possible variable type would be a str

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+
+    Returns
+    -------
+    list
+        Filtered list of discoveries
+    """
+    if state == 'all':
```
marcorosa, line 224: if the state is not specified we export all the discoveries. If the state is set, we export only that state.
Let's not give the possibility to set state=all because "all" never appears in the definition of state

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,279 @@
...
+        console.print(f'[bold] {args.repo_url} has 0 discoveries.')
+    elif len(discoveries) > MAX_NUMBER_DISCOVERIES:
+        response = ''
+        while response.upper() not in ['Y', 'N', 'YES', 'NO']:
+            response = console.input(
+                f'[bold]This repository has more than {MAX_NUMBER_DISCOVERIES} \
+discoveries, export them as .csv instead? (Y/N) ')
+        if response.upper() in ['N', 'NO']:
+            print_discoveries(discoveries, args.repo_url)
+        else:
```
marcorosa, line 272: so if someone types 'Z' shall we set it to "yes"?

marcorosa (changes requested):

credentialdigger/cli/get_discoveries.py
alaabenfatma: No, if someone types 'Z', he will be asked to enter another input in ['Y', 'N', 'YES', 'NO'] and will not be able to leave the **while** loop.

When testing `if response.upper() in ['N', 'NO']:`, we know that the input will 100% be in the list ['Y', 'N', 'YES', 'NO']. so if it is not 'N' nor 'NO', then it'll be 'Y' or 'YES'.

credentialdigger/cli/get_discoveries.py
alaabenfatma: We ask the user to provide the CLI with a path if the number of discoveries is 30 and they enter 'Y' (to export the discoveries).

If --save is specified (not False), we won't show the message requesting the user to specify the path and we will just export the discoveries directly.

If we happen to access export_csv with save == False, then this means that the repo has more than 30 discoveries and the user wants to export them instead of printing them on the Terminal.

credentialdigger/cli/get_discoveries.py
alaabenfatma: To be added to the licenses file.

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,276 @@
...
+def filter_discoveries(discoveries, state=None):
+    """ Filter discoveries based on state
+
+    Parameters
+    ----------
+    discoveries: list
+        List of discoveries to be filtered
+    states: str
+        str: if it is not specified, then return all discoveries.
+             return chosen state otherwise (i.e 'false_positive').
```
marcorosa, line 214: why is this line indented?

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,276 @@
...
+
+    Returns
+    -------
+    list
+        Filtered list of discoveries
+    """
+    if state is None:
+        state = ['new', 'false_positive',
+                 'addressing', 'not_relevant', 'fixed']
```
marcorosa, line 223:
```
if not state:
    return discoveries
return list(filter(lambda d: d.get('state') == state, discoveries))
```

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,276 @@
...
+
+    try:
+        csv_file = open(path, newline='', mode='w')
+    except IOError as e:
+        console.print(f'[red]{e}[/]')
+    else:
+        with csv_file:
+            with console.status('[bold]Exporting the discoveries..'):
+                # Add the category to each discovery
+                assign_categories(client, discoveries)
```
marcorosa, line 178: I suggest not to assign the category, at the moment. Maybe we'll introduce it later, but since it's not even in the `discoveries` table I think we can avoid it.

credentialdigger/cli/get_discoveries.py
```diff
@@ -0,0 +1,276 @@
...
+        while path == '':
+            path = console.input('Path to export CSV:')
+    else:
+        # if --save argument is set, we use it as an export path
+        path = save
+
+    try:
+        csv_file = open(path, newline='', mode='w')
+    except IOError as e:
+        console.print(f'[red]{e}[/]')
```
marcorosa, line 173: After the error message, isn't it a good idea to also print a message saying that the file can't be created and therefore the discoveries can't be exported?

marcorosa (changes requested):

marcorosa:
Please fix also the points raised by @pep8speaks

marcorosa (approved):

PR #171, opened 2021-08-11
Update base.html

sarthakg1234:
use defer for parallel loading


marcorosa:
As agreed, I close this PR together with #172 #173  #174 #175 #176 because the issue described in #177 is not due to this

PR #178, opened 2021-08-12
[Quick fix] Correct postgres operator.

alaabenfatma:
Fixed wrong prostgres SQL query.


marcorosa (approved):
the `LIKE` here wasn't an error, but I agree that having a `=` is better, both for clarity, performance, and because the states come from an enum type and are all different, so we don't have any advantage in using `LIKE`.

PR #179, opened 2021-08-12
[Feat + Fix] Faster calculations and retrieval of data + Fix breaking non-scalable UI

alaabenfatma:
### Introduction
When the user scans a lot of repositories, they fill-up the database with multiple new entries. 
When the database grows in size, collecting data becomes a lengthy task and blocks the UI (by keeping the other SQL queries pending) resulting in a non-functional UI (home page).

### Fix
We propose doing the calculations of the metadata with a multi-purpose SQL query that returns the following:

- Number of discoveries per repo
- Number of **true positive** discoveries per repo (part of the total number of the discoveries calculated previously)

### Performance
| Database | Old Time     | New Time       |
|----------|--------------|----------------|
| SQLite   | ~ 14 seconds | ~ 2 seconds    |
| Postgres | ~ 7 seconds  | ~ 0.05 seconds |

### DIFF
--- a/ui/backend/client_ui.py
+++ b/ui/backend/client_ui.py
@@ -49,8 +49,8 @@ class UiClient(Client):
         """
         query = '''SELECT repo_url,
                           COUNT(*) as total,
-                          sum(case when STATE='new' then 1 else 0 end) as tp 
-                          FROM discoveries 
+                          sum(case when STATE='new' then 1 else 0 end) as tp
+                          FROM discoveries
                           GROUP BY repo_url;
                 '''
         cursor = self.db.cursor()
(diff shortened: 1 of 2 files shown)


marcorosa:
As a suggestion for future PRs/analyses: if you want to compare performance you should give an indication of the setting you're using. In this specific case, for example, you manage to pass from 14 seconds to 2 seconds. But what's the setting? 14s vs 2s it's not a constant time, it strongly depends on the number of repos and discoveries together with the hardware you're running the UI on (a lag when you have 1GB RAM is not the same of a lag when you have 16GB RAM, and I could tell the same for the internet connection in case you use a remote server).

By the way the PR is good and I will merge it soon.

PR #186, opened 2021-08-18
[Feat] Export discoveries from the UI as a CSV file (faster version)

alaabenfatma:
This PR targets this issue: https://github.com/SAP/credential-digger/issues/114

It also hones the **get_all_discoveries_count** function, so that it retrieves all the discoveries counts at once.

### DIFF
--- a/ui/server.py
+++ b/ui/server.py
@@ -316,11 +316,22 @@ def get_repos():
             if repo['url'] == metadata[0]:
                 repo['total'] = metadata[1]  # Total number of discoveries
                 repo['TP'] = metadata[2]  # Number of new discoveries
+                repo['FP'] = metadata[3]  # Number of true positive discoveries
+                # Number of addressing discoveries
+                repo['addressing'] = metadata[4]
+                # Number of irrelevant discoveries
+                repo['not_relevant'] = metadata[5]
+                repo['fixed'] = metadata[6]  # Number of fixed discoveries
                 break  # We found the repo, no need to check next metadata
             else:
                 # The repo was scanned and doesn't have discoveries
                 repo['total'] = 0
                 repo['TP'] = 0
+                repo['FP'] = 0
+                repo['addressing'] = 0
+                repo['not_relevant'] = 0
+                repo['fixed'] = 0
+
         repo['scan_active'] = False
         if repo['url'] in active_scans:
             repo['scan_active'] = True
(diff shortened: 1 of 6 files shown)


ui/server.py
```diff
@@ -316,18 +318,63 @@ def get_repos():
...
+                  'addressing', 'not_relevant', 'fixed']
+    else:
+        states = request.form.getlist('check')
+
+    filtered_discoveries = list(
+        filter(lambda d: d.get('state') in states, discoveries))
+
+    try:
+        stringIO = io.StringIO()
+        csv_writer = csv.DictWriter(stringIO, discoveries[0].keys())
```
marcorosa, line 361: do you get an IndexError if you try to export discoveries for a repo that didn't have any?

ui/server.py
```diff
@@ -316,18 +318,63 @@ def get_repos():
...
+                repo['fixed'] = metadata[6]  # Number of fixed discoveries
                 break  # We found the repo, no need to check next metadata
             else:
                 # The repo was scanned and doesn't have discoveries
                 repo['total'] = 0
                 repo['TP'] = 0
+                repo['FP'] = 0
+                repo['addressing'] = 0
+                repo['not_relevant'] = 0
+                repo['fixed'] = 0
```
marcorosa, line 335: Are these items re-assigned at every loop? Maybe it's better to move them above the `for`, as an initialisation for the repo

marcorosa (changes requested):

ui/server.py
alaabenfatma: Yes.

ui/server.py
alaabenfatma: Do you suggest having the initialization of the fields before traversing the list of metadata, like this?
```python
    for repo in repos:
        # Init repo
        repo['total'] = 0
        repo['TP'] = 0
        repo['FP'] = 0
        repo['addressing'] = 0
        repo['not_relevant'] = 0
        repo['fixed'] = 0
        for metadata in repos_metadata:
            if repo['url'] == metadata[0]:
                repo['total'] = metadata[1]  # Total number of discoveries
                repo['TP'] = metadata[2]  # Number of new discoveries
                repo['FP'] = metadata[3]  # Number of true positive discoveries
                # Number of addressing discoveries
                repo['addressing'] = metadata[4]
                # Number of irrelevant discoveries
                repo['not_relevant'] = metadata[5]
                repo['fixed'] = metadata[6]  # Number of fixed discoveries
                break  # We found the repo, no need to check next metadata
```

marcorosa: yes, I think it's better so to not reassign the same values at every loop, even if the result doesn't change

marcorosa (approved):

PR #187, opened 2021-08-19
[Feat] Replace the use of TQDM with Rich

alaabenfatma:
[Rich](https://github.com/willmcgugan/rich) has been introduced into the project via this PR: https://github.com/SAP/credential-digger/pull/164

This PR removes TQDM and uses Rich instead in order to show scanning progress for the user. Under normal circumstances, we should get the following progress bar:
image

Note: tqdm is still used in the generator.py + qlearning.py files. I did not update those files since we may delete them in the future.

### DIFF
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -1144,12 +1144,11 @@ class Client(Interface):
         discoveries_ids = list()
         if debug:
             logger.debug('Update database with these discoveries.')
-            total_discoveries = len(new_discoveries)
             with Progress() as progress:
                 inserting_task = progress.add_task('Inserting discoveries...',
-                                                   total=total_discoveries)
-                for i in range(total_discoveries):
-                    curr_d = new_discoveries[i]
+                                                   total=len(new_discoveries))
+                for curr_discovery in new_discoveries:
+                    curr_d = curr_discovery
                     new_id = self.add_discovery(
                         curr_d['file_name'], curr_d['commit_id'],
                         curr_d['line_number'], curr_d['snippet'], repo_url,
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -1199,7 +1199,7 @@ class Client(Interface):
             total_discoveries = len(discoveries)
             false_positives = 0
             with Progress() as progress:
-                scanning_task = progress.add_task("Scanning discoveries...",
+                scanning_task = progress.add_task('Scanning discoveries...',
                                                   total=total_discoveries)
                 for i in range(total_discoveries):
                     false_positives += _analyze_discovery(discoveries[i])
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -1146,7 +1146,7 @@ class Client(Interface):
             logger.debug('Update database with these discoveries.')
             total_discoveries = len(new_discoveries)
             with Progress() as progress:
-                inserting_task = progress.add_task("Inserting discoveries...",
+                inserting_task = progress.add_task('Inserting discoveries...',
                                                    total=total_discoveries)
                 for i in range(total_discoveries):
                     curr_d = new_discoveries[i]
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -1147,8 +1147,7 @@ class Client(Interface):
             with Progress() as progress:
                 inserting_task = progress.add_task('Inserting discoveries...',
                                                    total=len(new_discoveries))
-                for curr_discovery in new_discoveries:
-                    curr_d = curr_discovery
+                for curr_d in new_discoveries:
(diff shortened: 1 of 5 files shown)


credentialdigger/client.py
```diff
@@ -1144,14 +1144,19 @@ def _scan(self, repo_url, scanner, models=None, force=False, debug=False,
...
-                curr_d = new_discoveries[i]
-                new_id = self.add_discovery(
-                    curr_d['file_name'], curr_d['commit_id'],
-                    curr_d['line_number'], curr_d['snippet'], repo_url,
-                    curr_d['rule_id'], curr_d['state'])
-                if new_id != -1 and curr_d['state'] != 'false_positive':
-                    discoveries_ids.append(new_id)
+            total_discoveries = len(new_discoveries)
+            with Progress() as progress:
+                inserting_task = progress.add_task("Inserting discoveries...",
```
marcorosa, line 1149: please fix string with `'` instead of `"` (consistency with the rest of the codebase)

credentialdigger/client.py
```diff
@@ -1191,11 +1196,14 @@ def _analyze_discovery(d):
...
             model_name = model_manager.model.__class__.__name__
             logger.debug(f'Analyzing discoveries with model {model_name}')
-
+            total_discoveries = len(discoveries)
             false_positives = 0
-            for i in tqdm(range(len(discoveries))):
-                false_positives += _analyze_discovery(discoveries[i])
-
+            with Progress() as progress:
+                scanning_task = progress.add_task("Scanning discoveries...",
```
marcorosa, line 1202: please fix string with `'` instead of `"` (consistency with the rest of the codebase)

credentialdigger/client.py
```diff
@@ -1144,14 +1144,19 @@ def _scan(self, repo_url, scanner, models=None, force=False, debug=False,
...
-                    curr_d['file_name'], curr_d['commit_id'],
-                    curr_d['line_number'], curr_d['snippet'], repo_url,
-                    curr_d['rule_id'], curr_d['state'])
-                if new_id != -1 and curr_d['state'] != 'false_positive':
-                    discoveries_ids.append(new_id)
+            total_discoveries = len(new_discoveries)
+            with Progress() as progress:
+                inserting_task = progress.add_task("Inserting discoveries...",
+                                                   total=total_discoveries)
+                for i in range(total_discoveries):
```
marcorosa, line 1151: if we replace tqdm then we can also update this loop into a for-each (`for curr_discovery in new_discoveries`) since we don't need the index `i` anymore

credentialdigger/client.py
```diff
@@ -1191,11 +1196,14 @@ def _analyze_discovery(d):
...
-
+            total_discoveries = len(discoveries)
             false_positives = 0
-            for i in tqdm(range(len(discoveries))):
-                false_positives += _analyze_discovery(discoveries[i])
-
+            with Progress() as progress:
+                scanning_task = progress.add_task("Scanning discoveries...",
+                                                  total=total_discoveries)
+                for i in range(total_discoveries):
```
marcorosa, line 1204: again, we don't need the index anymore so we can loop directly on the discoveries

marcorosa (changes requested):
I have a couple of requests for this file, but I also have a further consideration:
If we want to deprecate `tqdm` in favour of the use of `rich` then we must replace all the occurrences. Indeed, tqdm is still used in other files apart from client.py and we need to check if we can replace also those ones with rich. Finally, if we deprecate a dependency, we need to update `requirements.txt` and the third party notices accordingly.

credentialdigger/client.py
```diff
@@ -1144,14 +1144,18 @@ def _scan(self, repo_url, scanner, models=None, force=False, debug=False,
...
-                    curr_d['file_name'], curr_d['commit_id'],
-                    curr_d['line_number'], curr_d['snippet'], repo_url,
-                    curr_d['rule_id'], curr_d['state'])
-                if new_id != -1 and curr_d['state'] != 'false_positive':
-                    discoveries_ids.append(new_id)
+            with Progress() as progress:
+                inserting_task = progress.add_task('Inserting discoveries...',
+                                                   total=len(new_discoveries))
+                for curr_discovery in new_discoveries:
+                    curr_d = curr_discovery
```
marcorosa, line 1151: not needed... use the same variable! :D

marcorosa: if you don't want to update the references below, call it `curr_d` in the for loop...

marcorosa (changes requested):

credentialdigger/client.py
alaabenfatma: You are totally right, sorry what a noob mistake :)

marcorosa (approved):

PR #222, opened 2022-03-07
Bump/ubuntu22

marcorosa:
prepare the tool to be working with python3.10 and test it with ubuntu22.04


marcorosa:
Comment to myself: Slim added more test cases in the test repo, so discoveries number needs to be updated

PR #224, opened 2022-03-21
bug_fix update functions

LorisOnori:
- Add ```return``` keyword in: ```update_repo```, ```update_discovery```, ```update_discoveries```, ```update_discovery_group``` in ```client_sqlite.py``` and ```client_postgres.py``` so that those functions return a ```bool``` instead of ```None```. 

- Update of ```update_similar_snippet``` function so that the counter of how many updates has been made takes into account only successful updates.

- Change of ```query_check``` in client_sqlite.py to return True if changes has been made.

### DIFF
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -1366,7 +1366,7 @@ class Client(Interface):
             # Compute similarity of target_embedding and embedding
             similarity = compute_similarity(target_embedding,
                                             embedding)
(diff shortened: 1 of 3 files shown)


marcorosa:
Hi @LorisOnori please fix the code style as requested by our bot @pep8speaks 
Then I will review it

marcorosa (approved):

Issue #183, opened 2021-08-17
scan_user fails when clone_repo fails


marcorosa:
`scan_user` may break when some repository can't be cloned

Example:
```
INFO:credentialdigger.client:1/n) Scanning https://api.github.com/repos/USER/REPO
DEBUG:credentialdigger.client:Start scan
DEBUG:credentialdigger.scanners.git_scanner:Authenticate user with token
Traceback (most recent call last):
...
  File "/Users/i355397/git/credential-digger/venv/lib/python3.9/site-packages/credentialdigger-4.0.4-py3.9.egg/credentialdigger/cli/scan_user.py", line 103, in run
    discoveries = client.scan_user(
  File "/Users/i355397/git/credential-digger/venv/lib/python3.9/site-packages/credentialdigger-4.0.4-py3.9.egg/credentialdigger/client.py", line 982, in scan_user
    missing_ids[repo_url] = self._scan(repo_url, scanner,
  File "/Users/i355397/git/credential-digger/venv/lib/python3.9/site-packages/credentialdigger-4.0.4-py3.9.egg/credentialdigger/client.py", line 1096, in _scan
    raise e
  File "/Users/i355397/git/credential-digger/venv/lib/python3.9/site-packages/credentialdigger-4.0.4-py3.9.egg/credentialdigger/client.py", line 1088, in _scan
    new_discoveries = scanner.scan(repo_url,
  File "/Users/i355397/git/credential-digger/venv/lib/python3.9/site-packages/credentialdigger-4.0.4-py3.9.egg/credentialdigger/scanners/git_scanner.py", line 143, in scan
    project_path, repo = self.get_git_repo(repo_url, local_repo)
  File "/Users/i355397/git/credential-digger/venv/lib/python3.9/site-packages/credentialdigger-4.0.4-py3.9.egg/credentialdigger/scanners/git_scanner.py", line 104, in get_git_repo
    raise e
  File "/Users/i355397/git/credential-digger/venv/lib/python3.9/site-packages/credentialdigger-4.0.4-py3.9.egg/credentialdigger/scanners/git_scanner.py", line 100, in get_git_repo
    GitRepo.clone_from(repo_url, project_path)
  File "/usr/local/lib/python3.9/site-packages/git/repo/base.py", line 1032, in clone_from
    return cls._clone(git, url, to_path, GitCmdObjectDB, progress, multi_options, **kwargs)
[...]

PR #225, opened 2022-03-23
catch clone_repo exception

LorisOnori:
Added ```GitCommandError``` exception handler in ```scan_user```. The exception is raised when the repository cannot be cloned. This pull request solves issue #183

### DIFF
--- a/credentialdigger/client.py
+++ b/credentialdigger/client.py
@@ -8,6 +8,7 @@ from datetime import datetime, timezone
 
 import yaml
 from github import Github
+from git import GitCommandError
 from rich.progress import Progress
 
 from .models.model_manager import ModelManager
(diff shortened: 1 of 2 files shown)


credentialdigger/client.py
```diff
@@ -8,6 +8,7 @@
 
 import yaml
 from github import Github
+from git import GitCommandError
```
marcorosa, line 11: Please follow alphabetical order of imports (you can use [`isort`](https://pycqa.github.io/isort/) if you don't want to do it manually)

credentialdigger/client.py
```diff
@@ -982,11 +983,16 @@ def scan_user(self, username, category=None, models=None, debug=False,
...
-                                               git_token=git_token)
+            try:
+                missing_ids[repo_url] = self._scan(repo_url, scanner,
+                                                   models=models,
+                                                   debug=debug,
+                                                   similarity=similarity,
+                                                   git_token=git_token)
+            except GitCommandError:
+                logger.info(f"{i}/{repos_num} Ignore {repo_url} "
+                            "(it can not be cloned)")
```
marcorosa, line 994: please note all the strings are defined by single quotes (`'`) and not double ones (`"`) so we have to be consistent

marcorosa: the logger is in an exception so it shouldn't be at info level.
Better to use either warning or error

credentialdigger/scanners/git_scanner.py
```diff
@@ -100,6 +100,7 @@ def get_git_repo(self, repo_url, local_repo):
                 GitRepo.clone_from(repo_url, project_path)
                 repo = GitRepo(project_path)
             except GitCommandError as e:
+                logger.debug("Repo can not be cloned")
```
marcorosa, line 103: same comments as before, on quotes and logger level

marcorosa (changes requested):
Hi @LorisOnori  I have some requests before merging your pull request.

Moreover, please add a reference to the issue you're solving

marcorosa (approved):

PR #232, opened 2022-05-04
Pre commit hook

LorisOnori:


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+-   id: credential-digger-hook
+    name: credential digger hook
+    description: This hook run credential digger on pre-commited files
```
marcorosa, line 3: Please capitalize Credential Digger

credentialdigger/cli/cli.py
```diff
@@ -95,6 +110,12 @@ def main(sys_argv):
         parents=[parser_dotenv, parser_sqlite])
     get_discoveries.configure_parser(parser_get_discoveries)
 
+    # hook subparser configuration
+    parser_get_discoveries = subparsers.add_parser(
+        'hook', help='Launch credential digger as a pre-commit hook',
```
marcorosa, line 115: Please capitalize Credential Digger

credentialdigger/cli/hook.py
```diff
@@ -0,0 +1,180 @@
...
+
+optional arguments:
+  -h, --help         show this help message and exit
+  --dotenv DOTENV    The path to the .env file which will be used in all
+                     commands. If not specified, the one in the current
+                     directory will be used (if present).
+  --rules RULES      Specify the yaml file path containing the scan rules
+                     e.g., /path/to/rules.yaml
+  --db_path DB_PATH  Specify the database file path where to save the results
+                     e.g., /path/to/data.db
```
marcorosa, line 19: can't we use `--sqlite` instead of `--db_path` to be consistent to the other CLIs? You did the same with `--dotenv`

credentialdigger/cli/hook.py
```diff
@@ -0,0 +1,180 @@
...
+    # subprocess.check_output return the output of the command i.e., userinpput
+    user_input = subprocess.check_output('read -p \"\" userinput && echo '
+                                         '\"$userinput\"',
+                                         shell=True, stdin=sys.stdin).rstrip()
+
+    return user_input.decode('utf-8')
+
+
+def run(args):
+    """Run credential digger on staged files
```
marcorosa, line 79: Please capitalize Credential Digger

credentialdigger/cli/hook.py
```diff
@@ -0,0 +1,180 @@
...
+
+    Parameters
+    ----------
+    args: `argparse.Namespace`
+        Arguments from command line parser.
+
+    Returns
+    -------
+        It exits with success code 0 if there are no discoveries, otherwise
+        it exits with an error code != 0 printing the leaks
```
marcorosa, line 89: We have to be consistent with the return type. Do we return an int or an exit code? we cannot return an int (0) in a case and an exit code in another case. You can check the following example from the `scan` CLI:

```
While this function returns nothing of use to the scanner itself, it
gives an exit status (integer) that is equal to the number of
discoveries. If it exits with a value that is equal to 0, then it means
that the scan detected no leaks in this repo.
```

Maybe it's what happens also in this new CLI, but in this case you have to reformulate this docstring

credentialdigger/cli/hook.py
```diff
@@ -0,0 +1,180 @@
...
+
+    files_status = system('git', 'diff', '--name-status', '--staged'
+                          ).decode('utf-8').splitlines()
+    files = []
+    for fs in files_status:
+        stats = fs.split()
+        status = stats[0]
+        # Takes the last filename which is, in case of renamed files,
+        # the new name
+        filename = stats[-1]
```
marcorosa, line 100: Can we have more than 1 filename?

credentialdigger/cli/hook.py
```diff
@@ -0,0 +1,180 @@
...
+        # D = deleted files
+        # R = renamed files
+        # It considers the first char because the renamed files are displayed
+        # as Rxxx where xxx is a number
```
marcorosa, line 104: Does `status` have a limited number of possible values? is it only for renamed files that the format is Rxxx or also for others (e.g., Dxxxx)

credentialdigger/cli/hook.py
```diff
@@ -0,0 +1,180 @@
...
+        if status[0] not in 'DR':
+            files.append(filename)
+
+    if args.db_path:
+        db_path = args.db_path
```
marcorosa, line 109: in case you change the arg name, remember to update it also here

credentialdigger/cli/hook.py
```diff
@@ -0,0 +1,180 @@
...
+    else:
+        db_path = str(Path.home() / '.local' / 'data.db')
+
+    c = SqliteClient(path=db_path)
```
marcorosa, line 113: what if the user passes the `dotenv` argument? Does it get ignored?

credentialdigger/cli/hook.py
```diff
@@ -0,0 +1,180 @@
...
+
+    new_discoveries = []
+    subprocess.run(f'echo \"\nChecking files={files} \" > /dev/tty',
+                   shell=True,
+                   stdout=subprocess.PIPE)
+
+    # For optimization purposes, the PathModel and the PasswordModel are
+    # separated, otherwise scan_path will call both models for each file
+    # With this implementation the discoveries are accumulated and the
+    # PasswordModel will be used only once for the password discoveries
```
marcorosa, line 128: *will be run only once and for all the password discoveries

credentialdigger/cli/hook.py
```diff
@@ -0,0 +1,180 @@
...
+    for file in files:
```
marcorosa, line 129: don't use `file` since it's a reserved keyword in python. Better call it `staged_file`

credentialdigger/cli/hook.py
```diff
@@ -0,0 +1,180 @@
...
+    if password_discoveries:
+        mm = ModelManager('PasswordModel')
+        disc = c._analyze_discoveries(mm, password_discoveries, debug=False)
+
+    list_of_discoveries = []
+    for d in disc:
+        if d['state'] == 'new':
+            list_of_discoveries.append(d)
+
+    # If there are keys, token ...
```
marcorosa, line 162: `# There may be also discoveries other than passwords`

marcorosa (changes requested):

marcorosa (approved):

PR #282, opened 2023-11-14
Bump hyperscan dep

marcorosa:
Bump hyperscan dependency.
hyperscan v0.6 should work seamlessly on all python versions and should not require any hyperscan package to be installed upfront on the OS -> TODO: update README and Docker container image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,8 +58,8 @@ jobs:
         pip install . --user
 
     - name: Install tests dependencies
-       run: |
-         pip install -r ./tests/tests-requirements.txt
+      run: |
+        pip install -r ./tests/tests-requirements.txt
 
     - name: Run unit tests
       run: |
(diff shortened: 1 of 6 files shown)


marcorosa:
Tests failing for python3.9 as it's wheels are missing (and also sdist is missing)
[ref](https://github.com/darvid/python-hyperscan/issues/96)

marcorosa:
Using hyperscan v0.5.0 only for python3.9 version

In addition, as `python setup.py install` is outdated, refer to `pip install .` in the README and test workflow

.github/workflows/test.yml
```diff
@@ -53,14 +53,9 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-pip-
 
-    - name: Install python dependencies
-      run: |
-        pip install -r ./requirements.txt
-        pip install -r ./tests/tests-requirements.txt
-
```
marcorosa, line 61: tests requirements need to be reintroduced otherwise tests will fail due to missing deps

PR #342, opened 2025-03-14
Display discoveries if --no_interaction is used

st0yanDimitrov:
Added code to print discoveries in case --no_interaction is used for better observability.


st0yanDimitrov:
fixed user mapping

